PR #25, opened 2024-01-10
feat: add support for windows

bdraco:


bdraco:
looks like windows github runners are broken so will have to do this tomorrow

bdraco:
I typoed the windows version
windows-lastest

github apparently hangs on that 🤣

PR #41, opened 2024-04-27
feat: add cython models for AdvertisementData and BLEDevice

bdraco:


bdraco:
The overload happens to late unfortunately for this to work

PR #46, opened 2024-05-22
feat: speed up dispatching bleak callbacks

bdraco:


bdraco:
Fully covered and passing via Home Assistant's tests

PR #48, opened 2024-06-19
feat: add a fast cython init path for BluetoothServiceInfoBleak

bdraco:
This is ~20% speed up to processing raw adv data


bdraco:
additional code and complexity doesn't justify the speed up here

bdraco:
Probably makes more sense to avoid `_cython_init` completely and set the values one at a time

PR #116, opened 2025-01-22
feat: Add API for getting current slot allocations

bdraco:


src/habluetooth/manager.py
```diff
@@ -101,6 +101,7 @@ class BluetoothManager:
 
     __slots__ = (
         "_adapter_refresh_future",
+        "_adapter_sources",
```
bdraco, line 104:
```suggestion
        "_adapter_sources",
        "_allocations",
```

PR #147, opened 2025-02-04
fix: update poetry to v2

cdce8p:
Project metadata diff:
```diff
 ...
+License: Apache-2.0
-Requires-Python: >=3.11,<3.14
+Requires-Python: >=3.11
 ...
```


bdraco (approved):

pyproject.toml
```diff
@@ -1,18 +1,25 @@
-[tool.poetry]
+[project]
 name = "habluetooth"
 version = "3.21.0"
+license.text = "Apache-2.0"
```
cdce8p, line 4: We can't use just `license = "Apache-2.0"` yet since it isn't supported by setuptools. Same for the `dynamic` key, although that it seems to work fine even without it.

bdraco:
Thanks

PR #152, opened 2025-02-19
feat: fallback to bluetooth sock via management protocol when bluetoothd is hung

bdraco:
Since there is nothing we can do about bluetoothd being hung and not communicating with the container, use the bluetooth socket to start scans


bdraco:
Its a neat idea, the scan turns on but the dbus never gets updated.

PR #170, opened 2025-03-14
feat: improve performance for device with large manufacturer data history

bdraco:
We can short circuit the rich compare as soon as we know that the current adv is not a subset of the existing data since we know the current adv will always be small compared to the history.


bdraco:
sadly test_inject_100_different_manufacturer_data is the more common case so even if we can improve test_inject_100_different_advertisements its likely not worth it

PR #199, opened 2025-03-23
chore: compare rich compare overhead

bdraco:


bdraco:
still surprised the cython one I made is faster than CPython

PR #200, opened 2025-03-24
fix: project.license in pyproject.toml to allow package to build

EdwardBetts:
Fix for recent change to pyproject.toml that prevented the package from building.

```
configuration error: `project.license` must be valid exactly by one definition (2 matches found):
    - keys:
        'file': {type: string}
      required: ['file']
    - keys:
        'text': {type: string}
      required: ['text']
DESCRIPTION:
    `Project license <https://peps.python.org/pep-0621/#license>`_.
GIVEN VALUE:
    "Apache-2.0"
OFFENDING RULE: 'oneOf'
DEFINITION:
    {
        "oneOf": [
            {
                "properties": {
                    "file": {
                        "type": "string",
                        "$$description": [
                            "Relative path to the file (UTF-8) which contains the license for the",
                            "project."
                        ]
                    }
                },
                "required": [
                    "file"
                ]
            },
            {
                "properties": {
                    "text": {
                        "type": "string",
                        "$$description": [
                            "The license of the project whose meaning is that of the",
                            "`License field from the core metadata",
                            "<https://packaging.python.org/specifications/core-metadata/#license>`_."
                        ]
                    }
                },
                "required": [
                    "text"
                ]
            }
        ]
    }
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ build-backend = "poetry.core.masonry.api"
 [project]
 name = "habluetooth"
 version = "3.38.0"
-license = "Apache-2.0"
+license = {text = "Apache-2.0"}
 description = "High availability Bluetooth"
 authors = [{ name = "J. Nick Koston", email = "bluetooth@koston.org" }]
 readme = "README.md"


bdraco:
Cc @cdce8p

cdce8p:
Hey @EdwardBetts, may I ask what build command you used? The error suggests that at some point an old setuptools version was used, so maybe check for that as well. I'm a bit surprised though as #196 also updated the `build-system.requires` for setuptools to `>=77.0`.

_Also note that the lastest release `3.38.0` worked fine and includes that change._
https://github.com/Bluetooth-Devices/habluetooth/actions/runs/14010932498/job/39230664894
https://pypi.org/project/habluetooth/3.38.0/#files

EdwardBetts:
I see what went wrong. I'm trying to build with setuptools 75.8.0, this is the current version in Debian unstable.

I have packaged habluetooth for Debian and I was updating the package to the most recent release. This is the build command used:

```
python3.13 -m build --skip-dependency-check --no-isolation
```

Debian package of habluetooth: https://packages.debian.org/source/sid/habluetooth

I will wait until the required version of setuptools is in Debian before updating habluetooth package to the current release.

Feel free to close this pull request.

PR #209, opened 2025-04-17
feat: improve performance of _async_on_advertisement

bdraco:
This is only ~1% faster, but its slightly less code that wasn't reusable.


bdraco:
1% faster.. but slightly less code

PR #226, opened 2025-05-03
DNM: Test Cython 3.1 compat

bdraco:


bdraco:
Looks like it crashes on `async with self._start_stop_lock:` in `async_stop`

PR #275, opened 2025-08-12
fix: solve performance regression while connecting in Linux

bdraco:
When bleak put the backends behind the bleak scanner, all the backends now have to make a copy of discovered devices every time we connect. Since linux is already in the right forward we can bypass that and look at seen_devices directly. Since we already rely on scanner._backend this is no less safe than the other place we have to do it anyways.


bdraco:
Ok even stranger, both dbus paths and addresses are in that dict

PR #369, opened 2026-03-15
perf: skip raw advertisement parsing when data unchanged

bdraco:
## Summary
- When raw advertisement bytes match the previous advertisement for the same address, skip `parse_advertisement_data_bytes` and the merge logic entirely
- Reuse the previous parsed data (service_data, manufacturer_data, service_uuids, name, tx_power) in a new `BluetoothServiceInfoBleak` with updated rssi and time
- Since the data objects are reused by identity, the manager's dedup check (`is not` + `!=`) short-circuits on the identity check
- Updates `.pxd` Cython declarations with typed locals for the fast path

## Test plan
- [x] Existing tests pass
- [ ] Verify no regression in advertisement processing with ESPHome/remote scanners

### DIFF
--- a/src/habluetooth/base_scanner.pxd
+++ b/src/habluetooth/base_scanner.pxd
@@ -74,7 +74,7 @@ cdef class BaseHaScanner:
     @cython.locals(info=BluetoothServiceInfoBleak)
     cdef dict _build_discovered_device_timestamps(self)
 
-    @cython.locals(parsed=tuple, prev_info=BluetoothServiceInfoBleak, info=BluetoothServiceInfoBleak)
+    @cython.locals(parsed=tuple, prev_info=BluetoothServiceInfoBleak)
     cpdef void _async_on_raw_advertisement(
         self,
         str address,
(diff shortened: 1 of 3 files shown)


bdraco:
The fast path duplicates the entire `BluetoothServiceInfoBleak` construction from `_async_on_advertisement_internal`. We can avoid this by reusing the previous parsed data and letting `_async_on_advertisement_internal` handle it — the merge logic will see identical objects via `is` checks and short-circuit.

```python
def _async_on_raw_advertisement(self, address, rssi, raw, details, advertisement_monotonic_time):
    if (prev_info := self._previous_service_info.get(address)) is not None and prev_info.raw == raw:
        # Reuse previously parsed data — skip parse_advertisement_data_bytes
        self._async_on_advertisement_internal(
            address, rssi, prev_info.name, prev_info.service_uuids,
            prev_info.service_data, prev_info.manufacturer_data,
            prev_info.tx_power, details, advertisement_monotonic_time, raw,
        )
        return
    parsed = parse_advertisement_data_bytes(raw)
    self._async_on_advertisement_internal(
        address, rssi, parsed[0], parsed[1], parsed[2], parsed[3], parsed[4],
        details, advertisement_monotonic_time, raw,
    )
```

This eliminates the duplicated 13-field copy block entirely. The merge in `_async_on_advertisement_internal` will see `service_data is prev_info.service_data` → `True` (same objects passed back in) and skip all merge work. Combined with #368's `_adv_data_changed` flag, the manager will also skip dedup.

PR #407, opened 2026-05-14
feat: detect scanners stuck reporting zero free connection slots

### LINKED ISSUE #340: Bluetooth proxy connection slots can become permanently stuck, blocking all BLE connections until proxy reboot
During my work on the HA bluetooth stack, I encountered this very unintuitive and nasty bug which took a long time to figure out how to restore my system to working order. 

### Problem:
Bluetooth proxies can enter a stuck connection-slot state where all connection slots are reported as permanently occupied (slots=0/N free) despite no active connections. When this happens, Home Assistant is unable to connect to Bluetooth devices (e.g. Yale door locks) even though they are discovered, have good RSSI, and previously worked correctly.

The issue persisted across HA restarts, integration reloads, integration removals/re-additions, and can only be resolved by rebooting the affected Bluetooth proxy devices, which immediately restores normal operation.

It was difficult to trace down exactly which bluetooth proxies were causing the problems, as I suspected that it should be have been the proxies nearest the disconnected peripherals, but in my case it turned out to be my 3 Everything Presence Lite devices which are not the proxies closest to my affected devices (Yale door locks).

 All 3 proxies persistently had (failures=0) (in_progress=0) (slots=0/3 free) 

```
2025-12-16 20:31:47.806 INFO (MainThread) [habluetooth.wrappers] 
B0:44:9C:05:05:61 - LQHEW0001S: Found 4 connection path(s), preferred order: 
hci0 (FC:70:2E:0D:D7:BA) (RSSI=-81) (failures=6) (in_progress=3) (slots=3/5 free) (score=-87.09), 
tv-presence (08:A6:F7:5C:05:C6) (RSSI=-80) (failures=0) (in_progress=0) (slots=0/3 free) (score=-127), 
master-bathroom-presence-5c0574 (08:A6:F7:5C:05:76) (RSSI=-82) (failures=0) (in_progress=0) (slots=0/3 free) (score=-127), 
everything-presence-lite-931bec (08:A6:F7:93:1B:EE) (RSSI=-92) (failures=0) (in_progress=0) (slots=0/3 free) (score=-127)
```

### Solution:
1. Enabling `habluetooth.wrappers: debug` is required to identify the problematic proxies.
2. Hard reboot the proxies.
3. Bluetooth connections should be restored immediately

### Additional notes:

- HA 2025.12.2
- Supervisor 2025.12.3
- I currently have 10 BLE proxies enabled in my system, but it seems only 4 of them are ever listed as a valid connection path. Why?

bluetoothbot:
## What

Detect when a Bluetooth scanner (typically an ESPHome proxy) is stuck reporting zero free connection slots, and surface it as a warning + diagnostic instead of letting it silently break BLE connections.

## Why

Closes #340. Proxies can enter a state where they report all slots as occupied even though no devices are actually connected. Recovery requires a proxy reboot, but the symptom is hidden — the user has to enable debug logging on `habluetooth.wrappers` and grep INFO output for `slots=0/N free` to even know which proxy to power-cycle.

## How

- `BluetoothManager` now tracks `_allocations_zero_since` per source: the timestamp when a source first reported `free == 0` (cleared on any non-zero update or scanner unregister).
- The existing periodic `_async_check_unavailable` calls a new `_async_check_stuck_allocations` which logs a one-shot WARNING once a source has stayed at zero free slots for `STUCK_ALLOCATION_THRESHOLD_SECONDS` (30 min) — but only if habluetooth has no in-flight connections through that scanner, to avoid false positives for legitimately busy proxies.
- New `async_stuck_allocations()` public query and a `stuck_allocations` field in `async_diagnostics()` for HA's diagnostics download.
- `STUCK_ALLOCATION_THRESHOLD_SECONDS` is exposed in `__init__` for tuning/tests.
- `.pxd` updated to keep the cython build in sync.

## Testing

- 5 new tests covering: detection at threshold, warning fires once per stuck cycle, recovery clears state, in-flight connections suppress the warning, unregister cleans tracking, diagnostics surface stuck sources.
- Full suite: 230 passed, 1 skipped.


bdraco:
@bluetoothbot plan This proposed solution is the wrong fix. We should figure out why they are getting stuck in the first place

PR #445, opened 2026-05-22
test(benchmark): rename test_update_name_cache_* to test_seed_name_cache_*

bdraco:
Followup to #443 addressing the second of two late Copilot review comments.

The cdef `_update_name_cache` is not callable from Python; the microbenchmarks added in #443 actually measure through the `seed_name_cache` wrapper and include its call overhead. Rename the functions accordingly and refresh the section comment so the absolute numbers and what they represent are no longer misleading.

The first late comment (skip the helper call on passive ads for never-seen devices) was tried at a7ebd94 but regressed the dominant active-scanner-with-name path by ~9% on CodSpeed because the extra branches were paid on every advertisement for an optimization that only fires in a narrow window; that change was dropped.

### DIFF
--- a/tests/test_benchmark_base_scanner.py
+++ b/tests/test_benchmark_base_scanner.py
@@ -1026,11 +1026,17 @@ async def test_inject_100_bluez_raw_end_to_end_changed(
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bdraco:
@bluetoothbot review

PR #448, opened 2026-05-22
feat(manager): add AUTO scanning mode with on-demand active windows

bdraco:
Today every scanner runs in one fixed mode for its entire lifetime, so an integration that just needs an occasional 5s active sweep every few minutes forces the whole scanner into permanent active scanning. This adds a third BluetoothScanningMode value, AUTO, where scanners default to passive and the manager schedules short active windows on demand.

Callers declare a per-device active-scan need via BluetoothManager.async_register_active_scan(address, scan_interval, scan_duration). The scheduler indexes by address so the on_advertisement hot path is just two dict lookups; nothing is iterated when the advertisement does not match a registered address. Multiple registrations for the same address with different cadences coexist and fire on their own intervals, coalescing only when due in the same tick. scan_interval and scan_duration are validated at registration time. Each AUTO scanner also runs a 15s rediscovery sweep 10 minutes after it joins and every 12 hours thereafter, staggered (first-sweep offset wraps within the initial-sweep window) so concurrently-registered scanners do not all flip ACTIVE in the same second.

Each AUTO scanner gets one persistent worker task that sleeps on an asyncio.Event with a wait_for timeout until the next due event; new advertisements and registrations just set the wake event, so no task is allocated per dispatch.

BaseHaScanner grows a no-op async_request_active_window(duration); HaScanner implements it by stop/restarting the BleakScanner in ACTIVE mode under the existing _start_stop_lock with a scheduled return to the prior mode. On Linux the 4th-attempt fallback to PASSIVE is detected and reported as a failed window so the scheduler does not believe an active sweep ran; if the active restart itself raises ScannerStartError the scanner is brought back up in the underlying AUTO/passive mode rather than left stopped. Remote scanners override on their side; the bleak-esphome wiring to aioesphomeapi.bluetooth_scanner_set_mode lands in Bluetooth-Devices/bleak-esphome#343 and the aioshelly wiring in home-assistant-libs/aioshelly#1147. AUTO is opt in, existing ACTIVE/PASSIVE callers are unchanged.

The bleak callback registration path is unchanged; bleak itself has no cadence concept so the kwargs an earlier revision tried to add there were dropped. Home Assistant integrations reach the new API via async_register_callback kwargs in home-assistant/core#171806.

### DIFF
--- a/src/habluetooth/auto_scheduler.py
+++ b/src/habluetooth/auto_scheduler.py
@@ -106,6 +106,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import logging
+import math
 from typing import TYPE_CHECKING
 
 from .const import (
@@ -626,9 +627,20 @@ class AutoScanScheduler:
             worker.wake()
 
     def _coalesce_duration(self, entries: list[ActiveScanRequest]) -> float:
-        """Pick the max requested duration, clamped to the configured range."""
+        """
+        Pick the max requested duration, clamped to the configured range.
+
+        ``async_register_active_scan`` already rejects NaN / inf
+        inputs, but ``ActiveScanRequest`` can be constructed directly
+        (e.g., internal callers); the ``isfinite`` guard here keeps a
+        bad value from poisoning ``call_later``'s timeout.
+        """
         requested = max(
-            (e.scan_duration for e in entries if e.scan_duration is not None),
+            (
+                e.scan_duration
+                for e in entries
+                if e.scan_duration is not None and math.isfinite(e.scan_duration)
+            ),
             default=_AUTO_WINDOW_MIN_DURATION,
         )
         if requested < _AUTO_WINDOW_MIN_DURATION:
--- a/src/habluetooth/auto_scheduler.py
+++ b/src/habluetooth/auto_scheduler.py
@@ -471,10 +471,15 @@ class AutoScanScheduler:
         on_advertisement re-creates it the next time the device is
         seen.
 
-        Only wakes the owner's worker when this call actually inserted
-        a fresh entry into ``_needs``; re-registering an identical
-        request (e.g., from an HA config-entry reload) is a no-op on
-        the schedule so the wake would just churn ``_next_event_at``.
+        ``ActiveScanRequest`` is compared by identity, so each public
+        call to ``BluetoothManager.async_register_active_scan`` creates
+        a new request that contributes its own cadence to the same
+        address (two callers asking for windows every 60s on the same
+        device get two independent 60s cadences, not one). Adding the
+        *same* request object twice is idempotent and no-ops the
+        wake. Cancellation is per-registration — the callable returned
+        from ``async_register_active_scan`` only removes that specific
+        request, not other registrations against the same address.
 
         Pre-``start()`` registrations (no event loop yet) record the
         request only — no ``_needs`` entry is seeded. The first window
--- a/src/habluetooth/auto_scheduler.py
+++ b/src/habluetooth/auto_scheduler.py
@@ -432,11 +432,14 @@ class AutoScanScheduler:
 
         Sync to match ``BluetoothManager.async_stop``;
         ``worker.stop()`` calls ``task.cancel()`` without awaiting.
-        Cancellation lands on the next loop iteration and asyncio
-        reaps the task; a mid-``_tick`` scanner call may complete
-        first. Harmless for HA shutdown (scanners are being torn
-        down). Non-teardown callers should run the loop one more
-        iteration to know workers have actually stopped.
+        Cancellation lands on the next loop iteration; a mid-``_tick``
+        scanner call may complete first. Harmless for HA shutdown.
+        Callers doing an in-place restart (``stop()`` then ``start()``
+        on the same scheduler) must ``await asyncio.sleep(0)`` between
+        them so cancelled tasks finish their finally blocks before
+        new workers spawn on the same sources; ``start()`` does not
+        guard against this since HA's setup/teardown flow never
+        does an in-place restart.
         """
         self._running = False
         for worker in self._workers.values():
--- a/src/habluetooth/auto_scheduler.py
+++ b/src/habluetooth/auto_scheduler.py
@@ -153,6 +153,16 @@ class ActiveScanRequest:
 class _ScannerWorker:
     """One persistent task per AUTO scanner; sleeps until next due event."""
 
+    __slots__ = (
+        "_manager",
+        "_scanner",
+        "_scheduler",
+        "_sweep_last_completed",
+        "_task",
+        "_wake",
+        "_window_end",
+    )
+
     def __init__(
         self,
         scheduler: AutoScanScheduler,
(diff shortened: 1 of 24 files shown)


bdraco:
@bluetoothbot review

bdraco:
Pushed fixes for all 6 Copilot findings in d71c8ef:

- Hot path: track callbacks with scan_interval in a separate set so on_advertisement early-returns when none are registered. Should recover the codspeed regression.
- Busy loop: when _dispatch_per_device skips a busy scanner, defer the affected callbacks' due times past the window end so the next tick does not re-fire every 50ms.
- Failed window: capture the bool return; on False or exception, drop the _scanner_windows busy marker immediately and (for sweeps) leave _sweep_last_completed alone so the same scanner is retried.
- Shutdown: stop() now cancels every _pending_tasks entry, clears _scanner_windows, and resets _sweep_in_flight.
- Pre-start sweep timing: start() overwrites _sweep_last_completed instead of setdefault, so placeholders from pre-start add_scanner calls don't trigger an immediate first sweep.
- HaScanner.async_stop: also clears _scan_mode_override and _active_window_end so a later restart cannot come back in continuous ACTIVE.

Four new tests cover the hot-path early return, the busy-scanner deferral, the False-return path, and the stop-cancels-pending-tasks shutdown.

bdraco:
Pushed cfec48f: cythonize AutoScanScheduler so the per-advertisement dispatch from manager._scanner_adv_received goes through a direct vtable call instead of a Python attribute lookup plus method dispatch. This is the actual root cause of the codspeed regression; auto_scheduler.py is now in TO_CYTHONIZE and an auto_scheduler.pxd cdefs the class with cpdef versions of the methods called from cython code. Manager.pxd cimports it and types _auto_scheduler accordingly.

bdraco:
Addressed the bluetoothbot review across d54f0b4 and 4d3f3ba:

- 🔴 active fallback: async_request_active_window now checks current_mode after the swap and returns False on the Linux passive fallback path so the scheduler does not record a window that did not actually run
- 🟡 sweep backoff: _run_window advances _sweep_last_completed on failure too, preventing the 50ms tight retry loop when a sweep target persistently returns False
- 🟡 stale docstring: dropped the 'address and / or service UUID' wording that lingered after the address-only refactor
- 🟢 hot path: hoisted the _reschedule call out of the on_advertisement loop so multiple registrations for the same address only re-arm the tick timer once

Full suite green on the cython build, new test covers the sweep-backoff path.

src/habluetooth/scanner.py
```diff
@@ -626,9 +640,83 @@ async def async_stop(self) -> None:
...
+        async with self._start_stop_lock:
+            if self._active_window_handle is not None:
+                # A new window took over; let it own the override and timer.
+                return
+            self._scan_mode_override = None
+            if not self.scanning:
+                return
+            try:
+                await self._async_stop_scanner()
+                await self._async_start()
```
bdraco: duplicate code. could use a self._async_restart_scanner()

bluetoothbot: Worth noting that `_async_restart_scanner` isn't a drop-in replacement here — it does three things `_async_end_active_window` doesn't want:

1. **Conditional adapter reset** (`if self._start_time == self._last_detection or self.time_since_last_detection() > SCANNER_WATCHDOG_MULTIPLE: await self._async_reset_adapter(True)`). A scanner ending a 30s ACTIVE window may legitimately have a stale `_last_detection` if the device is quiet during that window — calling `_async_reset_adapter(True)` here would reset the adapter unnecessarily, causing a much longer outage than a simple stop+start.
2. **`exception` log level vs `warning`** — `_async_restart_scanner` uses `_LOGGER.exception` for failure, `_async_end_active_window` uses `_LOGGER.warning`. Probably wants to stay `warning` because end-of-window failure is recoverable (the watchdog will retry).
3. **Pre-restart guards** — `_async_end_active_window` has `if self._active_window_handle is not None: return` (new window took over) and `if not self.scanning: return` (scanner stopped). Those are end-of-window-specific.

The actual duplication is just the bare stop+start sequence (`_async_stop_then_start_under_lock`), which is two lines and already factored. A cleaner approach might be to refactor `_async_restart_scanner` to take a `reset_if_stale: bool` flag and reuse it from both sites, but that touches the watchdog path too — probably not worth it for two lines.

src/habluetooth/auto_scheduler.py
```diff
@@ -0,0 +1,604 @@
...
+        # the actual stop/start cost. Failure of the scanner call is
+        # handled the same way as success: we still don't retry until
+        # scan_interval out (or AUTO_REDISCOVERY_INTERVAL out for the
+        # sweep), which prevents busy-looping the worker on a stuck
+        # scanner.
+        self._advance_due(due_buckets, now)
+        if sweep_due:
+            self._sweep_last_completed = now
+        try:
+            await self._scanner.async_request_active_window(duration)
```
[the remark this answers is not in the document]
bdraco: no thats wrong

src/habluetooth/scanner.py
bluetoothbot: Three differences make a direct `self._async_restart_scanner()` reuse harder than it looks here:

1. **Lock semantics.** `_async_restart_scanner` does `async with self._start_stop_lock:` itself; `_async_stop_then_start_under_lock` is called from inside the `async with self._start_stop_lock:` block in `async_request_active_window` and `_async_end_active_window`. Reusing it would either double-acquire (deadlock with non-reentrant `asyncio.Lock`) or require restructuring both callers to release the lock first — which would reopen a race window where `_scan_mode_override` could be observed in transition.
2. **Adapter reset.** `_async_restart_scanner` reaches for `_async_reset_adapter(True)` when `_start_time == _last_detection` or the watchdog multiple has elapsed. That's the right behavior for a watchdog-driven "the radio has gone silent" path. For a mode flip on a healthy scanner it's the wrong thing — we'd reset hardware that's working fine just because the adapter happens to have been quiet during the last few seconds before we requested the window.
3. **Error propagation.** `_async_restart_scanner` catches `ScannerStartError` and logs it; the active-window path needs to *propagate* it so the surrounding `try/except ScannerStartError:` can drive the AUTO/passive recovery and return `False` to the scheduler.

A shared helper is feasible if you parameterize on (acquire-lock-or-not, do-adapter-reset-or-not, swallow-error-or-not) but that helper is essentially three booleans wrapping two lines (`await self._async_stop_scanner(); await self._async_start()`), so the abstraction would cost more than it saves. The duplication is two lines and the call sites read clearly with the differences explicit. Happy to extract if you want it, but I'd lean toward leaving as-is.

bluetoothbot: I'd push back on merging these into `_async_restart_scanner`. The two helpers have different contracts:

- `_async_restart_scanner` acquires `_start_stop_lock` itself, runs the adapter-reset conditional based on `_start_time == _last_detection` / `time_since_last_detection() > SCANNER_WATCHDOG_MULTIPLE`, and swallows `ScannerStartError` with `_LOGGER.exception`. It's the watchdog's entry point.
- `_async_stop_then_start_under_lock` is called from three lock-already-held sites in the new AUTO path: `async_request_active_window` (which needs to detect ACTIVE-start failure and explicitly fall back to passive), the ScannerStartError recovery within that same method (which needs to suppress a second start error), and `_async_end_active_window` (which logs the failure as a warning, not an exception, and only if the scanner was still supposed to be scanning).

If you swapped in `_async_restart_scanner` you'd either deadlock (it tries to acquire the lock the caller already holds) or have to refactor it to release/reacquire — which would invalidate every "caller holds lock" invariant in the AUTO path. You'd also pick up the adapter-reset and exception-swallow semantics that the AUTO callers specifically don't want (they need to surface failure to the scheduler so the override clears).

If the duplication still bothers you, the cleanest reuse direction is the other way: refactor `_async_restart_scanner` to call `_async_stop_then_start_under_lock` for its happy path, e.g.

```python
async def _async_restart_scanner(self) -> None:
    async with self._start_stop_lock:
        await self._async_stop_scanner()
        if (
            self._start_time == self._last_detection
            or self.time_since_last_detection() > SCANNER_WATCHDOG_MULTIPLE
        ):
            await self._async_reset_adapter(True)
        try:
            await self._async_start()
        except ScannerStartError as ex:
            _LOGGER.exception(...)
```

That keeps the watchdog semantics intact but lets the two-step pattern live in one place. Not necessary for this PR — just flagging the direction if you do want to deduplicate.

Issue #454, opened 2026-05-22
We should be using all the ruff rules that esphome/aioesphomeapi is using


bdraco:
We should be using all the ruff rules that esphome/aioesphomeapi is using

PR #477, opened 2026-05-23
ci: enable ruff TID (flake8-tidy-imports)

bdraco:
## Summary

Refs #454; aligns with the aioesphomeapi ruff config. Two relative imports in channels/bluez.py (`from ..const` and `from ..scanner`) become absolute (`from habluetooth.const` and `from habluetooth.scanner`).

## Test plan

- [x] `pre-commit run -a` green
- [x] `poetry run pytest tests/` green (390 pass, 1 skip)


bdraco:
lets skip this one

PR #483, opened 2026-05-23
ci: enable ruff PLR (pylint refactor)

bdraco:
## Summary

Refs #454; aligns with the aioesphomeapi ruff config. Three sites fixed after applying the same ignore list aioesphomeapi uses for the noisy `too-many-*` and `magic-value` checks (PLR0911, 0912, 0913, 0915, 2004).

PLR1730 (2): the if-compare-assign idiom collapses to `max(...)` in `AdvertisementTracker` and `min(...)` in `_ScannerWorker._next_event_at`; both auto fixes.

PLR1714 (1): the `self.name == "" or self.name == self.address` check in `BluetoothServiceInfoBleak._advertisement_internal` stays as is with a noqa comment. The mechanical `in ("", self.address)` rewrite makes cython emit an extra INCREF/DECREF pair on `self.name` (aliased into a temp for the membership check); this is the per advertisement hot path so the explicit short circuit pair of `==` is preferred.

Second of three PRs splitting the PL family.

## Test plan

- [x] `pre-commit run -a` green
- [x] `poetry run pytest tests/` green (390 pass, 1 skip)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -120,6 +120,11 @@ ignore = [
     "D401",    # First line of docstring should be in imperative mood
     "ASYNC109", # ``timeout`` is part of our bleak-compatible public API
     "TRY003",  # too many to fix; long messages outside the exception class
+    "PLR0911", # too many return statements (matches aioesphomeapi)
+    "PLR0912", # too many branches (matches aioesphomeapi)
+    "PLR0913", # too many arguments (matches aioesphomeapi)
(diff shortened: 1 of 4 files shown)


bdraco:
Skipping: PLR1714 made the per-advertisement hot path worse and the PLR1730 max()/min() fixes likely have the same overhead pattern in cython. Revisit if a cython-aware variant is added.

PR #551, opened 2026-06-22
refactor: split scanner into scanner_bleak with import shim

bdraco:
First step toward a dbus free local backend; this is a pure refactor with no behavior change.

The local bleak scanner moves to scanner_bleak.py. scanner.py stays as an alias that points habluetooth.scanner at the scanner_bleak module object via sys.modules, so both imports and patch("habluetooth.scanner.X") targets keep working unchanged for downstream consumers. The package exports, mgmt channel, and manager type hints now point at scanner_bleak, build_ext compiles the renamed module, and the in repo tests and example were repointed at scanner_bleak so mypy resolves them since the runtime alias is invisible to static analysis.

Verified green in both the pure python and compiled builds, and home-assistant tests/components/bluetooth passes unchanged.


bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed the downstream mock caveat from the review; instead of a plain re-export shim, scanner.py now aliases habluetooth.scanner to the scanner_bleak module object via sys.modules, so patch("habluetooth.scanner.X") resolves to the real code rather than a dead binding. Verified home-assistant tests/components/bluetooth passes unmodified (247 passed). In repo tests and the example were repointed at scanner_bleak so mypy can resolve them, and I added a test for the start loop fall-through to keep patch coverage clean.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Pinned the local scanner logger name back to habluetooth.scanner (it was resolving to habluetooth.scanner_bleak via __name__), so the move is fully behavior preserving including logging; anything configuring levels, filters, or log scraping on the habluetooth.scanner name keeps working. This also keeps the three caplog.at_level(logger="habluetooth.scanner") references in test_scanner.py correct, so no test changes were needed there. Verified green in pure python and compiled builds (582 passed).

bluetoothbot (approved):
No blocking issues found.

PR #554, opened 2026-06-22
feat: add mgmt discovery commands to bluez channel

bdraco:
Second step toward a dbus free local backend; this adds the management socket plumbing a future mgmt scanner will use to start and stop scanning without DBus.

MGMTBluetoothCtl gains start_discovery and stop_discovery for active scanning, add_adv_pattern_monitor and remove_adv_monitor for passive scanning, and a can_discover flag that is set once setup confirms NET_ADMIN/NET_RAW. Pending command futures are now keyed by (opcode, controller_idx) so the same command can be in flight on more than one adapter at once, and command complete handling was generalized to resolve any awaited command rather than just the capability probe. Nothing consumes these yet; the mgmt scanner lands in a follow up.

Green in pure python and compiled builds with new unit tests for every command and its failure paths.

### DIFF
--- a/src/habluetooth/channels/bluez.py
+++ b/src/habluetooth/channels/bluez.py
@@ -40,17 +40,39 @@ ADV_MONITOR_DEVICE_FOUND = 0x002F
 
 # Management commands
 MGMT_OP_GET_CONNECTIONS = 0x0015
+MGMT_OP_START_DISCOVERY = 0x0023
+MGMT_OP_STOP_DISCOVERY = 0x0024
 MGMT_OP_LOAD_CONN_PARAM = 0x0035
+MGMT_OP_ADD_ADV_PATTERNS_MONITOR = 0x0052
+MGMT_OP_REMOVE_ADV_MONITOR = 0x0053
 
 # Management events
 MGMT_EV_CMD_COMPLETE = 0x0001
 MGMT_EV_CMD_STATUS = 0x0002
+
+# Discovery address-type bitmask: BR/EDR (0x01) | LE Public (0x02) | LE Random (0x04).
+# We only scan for LE devices, so we combine the two LE bits.
+SCAN_TYPE_LE = 0x06
+
+# Management command status codes
+MGMT_STATUS_SUCCESS = 0x00
+# Busy means another mgmt client (e.g. bluetoothd) already started discovery;
+# device-found events are still delivered to every open socket, so we treat it
+# as success for our purposes.
+MGMT_STATUS_BUSY = 0x0A
 
 # Pre-compiled struct formats for performance
 COMMAND_HEADER = Struct("<HHH")
 COMMAND_HEADER_PACK = COMMAND_HEADER.pack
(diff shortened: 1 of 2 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- Wrong opcodes for advertisement-monitor commands (0x0029/0x0030 collide with existing commands)

bdraco:
Thanks, addressed all three.

Blocking opcode collision: fixed the advertisement monitor opcodes; ADD_ADV_PATTERNS_MONITOR is 0x0052 and REMOVE_ADV_MONITOR is 0x0053 per the kernel mgmt-api. I confirmed against btsocket that 0x0029 and 0x0030 are SetAdvertising and LoadIdentityResolvingKeys, and that 0x0023/0x0024 for start/stop discovery are correct.

Write error handling: _send_command_await now also catches BluetoothSocketError, which is not an OSError subclass, so a failing _write_to_socket is treated as a soft failure rather than propagating; added a parametrized test that exercises both OSError and BluetoothSocketError.

Import grouping: moved the TYPE_CHECKING only Callable import below the regular imports.

Green in pure python and compiled builds (596 passed).

bluetoothbot (changes requested):
Blocking issues found.

- stop_discovery coerces MGMT_STATUS_BUSY into "stopped"

bdraco:
Addressed the latest round.

stop_discovery no longer treats BUSY as stopped; it now returns True only on MGMT_STATUS_SUCCESS, since BUSY on stop means the kernel did not stop discovery (e.g. another client is still discovering). start_discovery keeps the SUCCESS or BUSY semantics, which is correct there. Added a test asserting stop with BUSY returns False.

Logging: a failed start_discovery now logs at warning since it means no advertisements, and a failed remove_adv_monitor logs at warning since it leaks a controller side monitor handle. add_adv_pattern_monitor now distinguishes an error status from a success response that is missing the 2-byte handle, warning separately for each; added a test for the missing-handle case.

Removed MGMT_EV_DISCOVERING since nothing references it yet; it will come back with the scanner that consumes it.

Green in pure python and compiled builds (598 passed), bluez.py has no uncovered lines in the diff.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Thanks for the approve. Folded in the non-blocking notes too.

Documented on command_response that only one command per (opcode, controller_idx) may be in flight at a time, since a second concurrent call with the same key would overwrite the first waiter; the wrappers never double-issue the same command on one adapter.

stop_discovery now logs the no-response case as well, for parity with start_discovery and remove_adv_monitor; it stays at debug since stop is best-effort cleanup on a shared socket.

Added a behavioral timeout test that drives start_discovery against a response that never arrives, so the asyncio_timeout path is exercised rather than just covered by the socket-error case.

Left _send_command_await at debug on the inner failure since the operation-level callers already warn on failure; bumping both would double-log during transient reconnects. Green in pure python and compiled (599 passed).

PR #556, opened 2026-06-22
refactor: move raw bluez advertisement adaptation into the bluez channel

bdraco:
Groundwork for the mgmt based local scanner; pure refactor with no behavior change.

The BlueZ specific adaptation (raw MAC byte order, org.bluez object path) moves out of the scanner classes and into the bluez channel. data_received now converts the address and builds the details, then calls the platform agnostic _async_on_raw_advertisement on whichever scanner registered for that controller. No scanner class carries a bluez method anymore, so BlueZ handling stays off the cross platform base and off the macOS local scanner entirely.

_side_channel_scanners and the channel are retyped from HaScanner to BaseHaScanner, which also drops a type ignore at the registration site. The bytes_mac_to_str and make_bluez_details helpers now live in the bluez channel where they are used.

Green in pure python and compiled builds, and home-assistant tests/components/bluetooth passes unchanged.

### DIFF
--- a/src/habluetooth/channels/bluez.pxd
+++ b/src/habluetooth/channels/bluez.pxd
@@ -1,7 +1,7 @@
 
 import cython
 
-from ..scanner_bleak cimport HaScanner
+from ..base_scanner cimport BaseHaScanner
 cdef bint TYPE_CHECKING
 
 cdef unsigned short DEVICE_FOUND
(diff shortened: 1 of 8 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Reworked per the review point that macOS is a local adapter with no BlueZ. Rather than putting the bluez handler on the base scanner (or a local scanner base, which would still be wrong for macOS), the bluez specific conversion now lives entirely in the bluez channel; data_received converts the address and builds the details and calls the generic _async_on_raw_advertisement. No scanner class carries a bluez method now, so the cross platform base and the macOS local scanner are both free of bluez code, and the upcoming mgmt scanner is fed by the same channel path without needing anything bluez specific itself.

bluetoothbot (approved):
No blocking issues found.

PR #559, opened 2026-06-23
feat: add l2cap att socket transport

bdraco:
Adds channels/l2cap.py, the raw L2CAP ATT socket transport for the kernel backend; it opens a connection oriented L2CAP socket to a peer on the ATT fixed channel (CID 0x0004) and exposes it as an asyncio push transport, one SEQPACKET datagram in, one ATT PDU out, which is what feeds the ATTClient codec.

CPython's socket module cannot express sockaddr_l2 (no l2_cid or l2_bdaddr_type), so the address is packed by hand and bind/connect go through libc via ctypes on the raw fd; everything after the connect handshake is ordinary asyncio, add_reader for the inbound stream and sock_sendall for backpressured writes. BT_SECURITY is requested before connect so the kernel drives LE encryption when bonded keys exist; pass security_level 0 to skip it.

Nothing imports this yet; it lands ahead of the GATT client. The libc bind/connect and the BT_SECURITY setsockopt are the only Bluetooth specific lines, marked as platform glue; the orchestration around them is exercised over an ordinary socket pair so the module is unit tested without hardware at 100 percent coverage.

### DIFF
--- a/src/habluetooth/channels/l2cap.py
+++ b/src/habluetooth/channels/l2cap.py
@@ -192,10 +192,14 @@ class L2CAPSocket:
         one fd (a second backpressured call would replace the first's writer and
         stall it). SEQPACKET keeps each write framed as one PDU.
         """
-        if self._closed:
-            msg = "L2CAP socket is closed"
-            raise BleakError(msg)
         async with self._write_lock:
+            # Re-check under the lock: close() may have run while we were parked
+            # on it (e.g. a peer disconnect tearing the socket down), and writing
+            # to the closed fd must raise the documented BleakError, not an
+            # incidental OSError from sock_sendall.
+            if self._closed:
+                msg = "L2CAP socket is closed"
+                raise BleakError(msg)
             await self._loop.sock_sendall(self._sock, data)
 
     def _read_ready(self) -> None:
--- a/src/habluetooth/channels/l2cap.py
+++ b/src/habluetooth/channels/l2cap.py
@@ -144,6 +144,7 @@ class L2CAPSocket:
         self._on_data = on_data
         self._on_close = on_close
         self._closed = False
+        self._write_lock = asyncio.Lock()
         loop.add_reader(sock.fileno(), self._read_ready)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- send() is not safe for the concurrent use ATTClient documents

bdraco:
Fixed both.

send() now serializes writes with an asyncio.Lock so it satisfies the concurrency contract ATTClient documents; sock_sendall is not safe to run concurrently on one fd (a second backpressured call replaces the first's writer and stalls it), and ATTClient sends indication confirmations and write-without-response commands outside its transaction lock. SEQPACKET still keeps each write framed as one PDU. Added test_send_serializes_concurrent_writes which gates a fake sendall to prove the second send waits for the first.

CI: the Windows jobs failed because the Windows default proactor event loop does not implement add_reader, add_writer, or sock_sendall, which this transport relies on; L2CAP is Linux only anyway. Skipped the module on win32 with a reason; the selector loops on Linux and macOS run the full suite, so coverage stays 100 percent.

l2cap.py 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Good catch; fixed. send() now re-checks _closed inside the write lock, so a caller parked on the lock while close() runs (e.g. a peer disconnect) raises the documented BleakError rather than an incidental OSError from sock_sendall. Added test_send_rechecks_closed_after_acquiring_lock which parks a second send on the lock, closes the socket, then asserts it raises closed.

100 percent line and branch, full suite green.

bluetoothbot (changes requested):
Blocking issues found.

- Write-path socket error doesn't tear down the channel (read/write teardown asymmetry)

bdraco:
Fixed. send() now wraps sock_sendall so a write-side OSError (ECONNRESET, EPIPE) routes through _fail(exc), which closes the socket and notifies on_close exactly once before re-raising to the caller, mirroring the read path and keeping the docstring contract deterministic. Added test_send_error_tears_down_channel which makes sendall raise and asserts the channel closes and on_close fires once with that error.

100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #560, opened 2026-06-23
feat: add l2cap gatt client backend

bdraco:
Adds client_mgmt.py, HaMgmtClient, a bleak BaseBleakClient that drives GATT over the raw L2CAP ATT channel instead of bluetoothd/DBus. It wires the L2CAPSocket transport to the ATTClient codec, runs MTU exchange and service discovery on connect, and translates the discovered tree into bleak's unified GATT model so existing integrations work unchanged; read, write with and without response, descriptor read/write, and notify/indicate via the CCCD are all implemented.

This is not wired into Home Assistant; scanner selection still builds the DBus backed HaScanner and bleak's platform client, so there is no behavior change. Like the codec and the transport, nothing imports this module yet; it is the connect side backend a later mgmt scanner and factory change will route to.

Pairing is intentionally out of scope here; pair and unpair raise and connect ignores the pair flag, the kernel still drives LE encryption from bonded keys via the socket BT_SECURITY level. Mgmt driven bonding lands in a follow up alongside the mgmt connect and pairing commands. The reconnect GATT cache is deferred to the scanner that owns it.

Tested by driving the real ATTClient against a fake transport wired to a small in memory GATT server, covering connect and discovery, the value handle mapping, read and write paths, notify and indicate subscription and routing, teardown on disconnect and on unexpected drop, and the not connected and pairing guards; 100 percent line and branch coverage, full suite green.

### DIFF
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -94,7 +94,16 @@ class HaMgmtClient(BaseBleakClient):
         self._adapter_address = client_data.adapter_address
         self._scanner = client_data.scanner
         details = getattr(address_or_ble_device, "details", None) or {}
-        self._address_type: int = details.get("address_type", BDADDR_LE_PUBLIC)
+        if "address_type" in details:
+            self._address_type: int = details["address_type"]
+        else:
+            # Adverts always carry the peer address type, so the scanner should
+            # populate it; fall back to public but log so a missing field is not
+            # only visible as an opaque L2CAP connect timeout later.
+            self._address_type = BDADDR_LE_PUBLIC
+            _LOGGER.debug(
+                "%s: no address_type in details; assuming LE public", self.address
+            )
         self._att: ATTClient | None = None
         self._sock: L2CAPSocket | None = None
         self._connected = False
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -170,6 +170,11 @@ class HaMgmtClient(BaseBleakClient):
 
     def _handle_disconnect(self, exc: Exception | None) -> None:
         """Tear the channel down once and notify on an unexpected drop."""
+        if exc is not None:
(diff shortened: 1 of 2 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed both.

1. Pinned the disconnect semantics: kept firing the disconnected callback on a deliberate disconnect to match the BlueZ backend this replaces, documented that in the disconnect docstring, and added test_disconnect_fires_disconnected_callback so a future refactor can not silently flip it.

2. start_notify now unwinds the notify handler if the CCCD write fails; kept the register before write ordering so a notification racing in right after the enable is not lost, and wrapped the write so a failure removes the handler before re-raising. Added test_start_notify_unwinds_handler_on_cccd_write_failure which fails the CCCD write and asserts a later stray notification is not delivered.

100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Done. The address_type fallback is now an explicit branch that debug logs when details has no address_type before assuming LE public, so a missing field is traceable at its source rather than surfacing as an opaque L2CAP connect timeout. Added test_address_type_falls_back_to_public_when_missing covering the fallback branch.

100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Done. _handle_disconnect now debug logs the originating cause (the codec passes the transport OSError or ATT timeout) before tearing down, gated on exc being present so a deliberate disconnect() that passes None stays silent. Now an encryption rejection, a peer reset, and a request timeout are distinguishable in the field instead of all looking like a clean disconnect. Added test_unexpected_drop_logs_cause asserting the cause surfaces in the log.

100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #561, opened 2026-06-23
feat: add mgmt-based local scanner and factory

bdraco:
Adds scanner_mgmt.py with HaScannerMgmt, a local scanner that drives discovery over the BlueZ management socket (active discovery or a passive advertisement monitor) and routes connections through the L2CAP HaMgmtClient, no DBus. Advertisements arrive over the existing mgmt side channel, which the manager points at this scanner once it is registered by adapter index, so the hot advert path is inherited unchanged from BaseHaScanner. Also adds create_local_scanner(mode, adapter, address), the factory that returns the mgmt scanner when mgmt discovery is available and falls back to the bleak backed HaScanner otherwise.

Marked experimental at the top of the module; the public names are likely to change or be renamed. Nothing in Home Assistant selects this scanner, the factory is not wired into core yet, so there is no behavior change; HA keeps using HaScanner and the bleak platform client. Verified the 26 habluetooth symbols HA imports all still resolve.

Connection slots are tracked by the scanner itself rather than BleakSlotManager; that manager counts connections by BlueZ DBus path, which an L2CAP connection never creates, so it cannot see mgmt connections. The scanner keeps its own set of live peer addresses, reads only the configured slot count from the slot manager as the limit, and the client reports connect and disconnect through two callbacks on MgmtClientData (additive, the existing client behavior is unchanged when they are absent).

Tested the factory selection, active and passive start, start failures, stop, the quiet watchdog restart and its skip and failure paths, teardown, and the slot counting and gating; scanner_mgmt and client_mgmt both at 100 percent line and branch, full suite green.

### DIFF
--- a/src/habluetooth/__init__.py
+++ b/src/habluetooth/__init__.py
@@ -30,6 +30,7 @@ from .models import (
 )
 from .scanner_bleak import BluetoothScanningMode, HaScanner, ScannerStartError
 from .scanner_device import BluetoothScannerDevice
+from .scanner_mgmt import HaScannerMgmt, create_local_scanner
 from .storage import (
     DiscoveredDeviceAdvertisementData,
     DiscoveredDeviceAdvertisementDataDict,
@@ -66,11 +67,13 @@ __all__ = [
     "HaBluetoothSlotAllocations",
     "HaScanner",
     "HaScannerDetails",
+    "HaScannerMgmt",
     "HaScannerModeChange",
     "HaScannerRegistration",
     "HaScannerRegistrationEvent",
     "HaScannerType",
     "ScannerStartError",
+    "create_local_scanner",
     "discovered_device_advertisement_data_from_dict",
     "discovered_device_advertisement_data_to_dict",
     "expire_stale_scanner_discovered_device_advertisement_data",
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -157,7 +157,14 @@ class HaMgmtClient(BaseBleakClient):
         self.services = self._build_services(services)
         self._connected = True
         if self._register_connection is not None:
-            self._register_connection(self.address)
+            # Slot bookkeeping is best-effort; a failure must not undo a
+            # connection that is already established.
+            try:
+                self._register_connection(self.address)
+            except Exception:
+                _LOGGER.exception(
+                    "%s: connection slot register callback failed", self.address
+                )
 
     async def disconnect(self) -> None:
         """
@@ -194,7 +201,15 @@ class HaMgmtClient(BaseBleakClient):
             self._sock = None
         if was_connected:
             if self._unregister_connection is not None:
-                self._unregister_connection(self.address)
+                # Best-effort: a bookkeeping failure must not stop the
+                # disconnected callback from firing.
+                try:
+                    self._unregister_connection(self.address)
+                except Exception:
+                    _LOGGER.exception(
+                        "%s: connection slot unregister callback failed",
+                        self.address,
+                    )
             if self._disconnected_callback is not None:
                 self._disconnected_callback()
 
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -76,6 +76,11 @@ class MgmtClientData:
 
     adapter_address: str  # local adapter BD_ADDR the L2CAP socket binds/sends from
     scanner: SupportsConnecting  # provides connecting() to pause scanning
+    # Optional slot bookkeeping: the mgmt scanner tracks live connections itself
+    # (BleakSlotManager is DBus-path based and cannot see L2CAP connections), so
+    # the client reports connect/disconnect by peer address here.
+    register_connection: Callable[[str], None] | None = None
+    unregister_connection: Callable[[str], None] | None = None
 
 
 class HaMgmtClient(BaseBleakClient):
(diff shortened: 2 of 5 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- Missing discovered-device read-out overrides — NotImplementedError once registered
- Connections bind by BD_ADDR while discovery is keyed by adapter index — they can diverge

bdraco:
Addressed all four.

1 (read-out overrides): HaScannerMgmt now overrides discovered_devices, discovered_devices_and_advertisement_data, discovered_addresses, and get_discovered_device_advertisement_data, served from the inherited _previous_service_info, so the manager's unavailable tracking, connect path, and diagnostics work once it is registered. I built the dicts from _previous_service_info directly rather than calling _build_discovered_device_advertisement_datas, since that helper is a private cdef method a pure-Python subclass cannot reach (same class of issue as the _manager fix). Added a test that registers the scanner, injects an advert, and reads all four; it exercises the cpdef override dispatch under the cython build.

2 (connect adapter divergence): create_local_scanner now also requires address != DEFAULT_ADDRESS, so the mgmt scanner is only chosen when there is a real BD_ADDR to bind the L2CAP socket to; a missing address would bind to BDADDR_ANY and let the kernel pick a different radio than the one discovering. Added a fallback test for the default-address case.

Silent failure 1 and 2: _async_stop_discovery now checks the bool returns; it only clears _monitor_handle when remove_adv_monitor succeeds (and keeps it to retry, with a warning, on failure), warns when stop_discovery fails, and debug logs when teardown is skipped because the controller went away. Added tests for the stop-failure and failed-removal branches.

Suggestion 1 (TOCTOU): documented the _can_connect gate as advisory, not a reservation, in its docstring.

Verified against a cython build (REQUIRE_CYTHON): scanner_mgmt and client_mgmt both 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed all three suggestions.

Backend name: the connector client is now a small functools.partial subclass named HaMgmtClient, so the connect path backend id (derived from type(connector.client).__name__) reads as HaMgmtClient instead of the generic partial. Asserted in the connector test.

Free slots: get_allocations now clamps free at 0 so a TOCTOU overshoot past the advisory gate cannot report negative free; the connection set is keyed by peer address (one connection per peer). Added an overshoot test.

Mode change: _async_start now uses self.set_current_mode(mode) instead of assigning current_mode directly, so mode-change subscribers are notified, matching HaScanner. Added a test asserting the manager is notified on start.

scanner_mgmt at 100 percent line and branch against a cython build, full suite green.

bdraco:
Addressed all five Copilot comments.

1 and 2 (slot callbacks can break connectivity): the register and unregister callbacks in HaMgmtClient are now best-effort; each is wrapped so a raising bookkeeping callback is logged via _LOGGER.exception and cannot undo an established connection (register) or stop the disconnected callback from firing (unregister). Added tests where each callback raises and the connection still establishes / still fires disconnect.

3 (direct construction with DEFAULT_ADDRESS): HaScannerMgmt.__init__ now raises ValueError on DEFAULT_ADDRESS rather than silently binding L2CAP to BDADDR_ANY and mis-routing through the wrong adapter; the factory already guarded this, now direct callers fail fast too. Added a test.

4 (AUTO mode): create_local_scanner now falls back to HaScanner for AUTO, since HaScannerMgmt does not implement async_request_active_window and so cannot promote the radio to active on demand; without the fallback AUTO semantics would silently differ for mgmt scanners. Added a test. A later PR that adds active-window support can drop this guard.

5 (allocations): free is already clamped at 0; allocated is now sorted for stable output so repeated reads do not look like allocation changes. Added a sorted-output test.

scanner_mgmt and client_mgmt both 100 percent line and branch against a cython build, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Done. The background-task done-callback now drops the reference and inspects the result: a non-cancelled task that finished with an exception is logged via _LOGGER.error with exc_info, so an unexpected escape (for example if _async_stop_discovery ever raised) surfaces in HA logs instead of only appearing as asyncio's Task exception was never retrieved at GC time. Added a test that runs a failing background task and asserts it is logged and the reference is dropped.

Scoped to HaScannerMgmt; left HaScanner's _create_background_task as is since that is the existing pattern and out of scope for this PR.

scanner_mgmt 100 percent line and branch, full suite green.

bluetoothbot (changes requested):
Blocking issues found.

- Restart clobbers an un-removed monitor handle, defeating the retry-and-don't-leak intent

bdraco:
Addressed the important item and the two related notes.

Important 1 (restart clobbers an un-removed monitor handle): fixed. _async_restart now bails out after _async_stop_discovery if _monitor_handle is still set, instead of falling through to _async_start which would add and overwrite-track a second monitor while the old one stays registered in the kernel. The watchdog re-fires next tick and retries the removal, so the documented retry-rather-than-leak path now actually fires on the restart flow. Added a test that fails the monitor removal, triggers the watchdog restart, and asserts no second monitor is added and the handle is preserved.

Suggestion 1 (restart catches only ScannerStartError, stop outside try): as the review notes, this is already observable (the done-callback logs escaped exceptions with exc_info) and self-healing (the watchdog re-arms every tick), and with the bail above a failed removal now defers cleanly to the next tick rather than stacking. The active path has no monitor and stop_discovery failures only warn, so I kept the focused ScannerStartError handler.

Silent failure (unregister leaves slot in _connections): the concrete unregister callback is HaScannerMgmt._unregister_connection, which is self._connections.discard(address); set.discard cannot raise, so the slot is never actually leaked. The try/except is purely defensive for arbitrary callbacks (added per the earlier Copilot review so a hostile callback can't block the disconnected callback); for our implementation the except branch is unreachable.

scanner_mgmt 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #562, opened 2026-06-23
feat: raise the maximum auto active scan window to 35s

bdraco:
Some scan response only sensors broadcast on a 30s interval (Govee H5074, Inkbird IBS-TH2), so a 30s active window can fall just short of a full interval and miss them. This raises AUTO_WINDOW_MAX_DURATION from 30s to 35s so callers can request a window that spans a 30s interval with margin.

The coalesce lookahead derives from the window max, so it stays ahead of it (now 40s); the invariant that lookahead exceeds the window is preserved.


bluetoothbot (approved):
No blocking issues found.

PR #563, opened 2026-06-23
feat: add mgmt pair, unpair, and disconnect commands

bdraco:
Adds three management socket command senders to the bluez channel; pair_device (bond with a peer, NoInputNoOutput IO capability so the kernel runs Just Works pairing and stores the keys), unpair_device (remove the bond, disconnecting by default), and disconnect_device (force-disconnect at the controller). Each follows the existing start_discovery / load_conn_params pattern, encodes the mgmt_addr_info layout (bdaddr reversed to little endian plus address type), awaits the command-complete status, and returns a bool.

Opcodes and the request struct layouts were taken from the installed btsocket btmgmt_protocol (PAIR_DEVICE 0x0019, UNPAIR_DEVICE 0x001B, DISCONNECT 0x0014; pair is addr+type+io_capability, unpair is addr+type+disconnect, disconnect is addr+type). A small helper packs and validates the address and treats a malformed address as a soft failure rather than raising into the command path.

This is the command sending half of the pairing work and unblocks HaMgmtClient.pair and unpair later; nothing calls these yet. The event side (NEW_LONG_TERM_KEY capture, AUTH_FAILED, key restore via LOAD_LONG_TERM_KEYS) and the passkey/confirmation interactive replies are follow ups. No behavior change.

Tested the encoded command bytes for each, success and failure status, the no response path, the unpair disconnect flag, and that a too short or non hex address is rejected without sending; the new code is fully covered.

### DIFF
--- a/src/habluetooth/channels/bluez.py
+++ b/src/habluetooth/channels/bluez.py
@@ -41,12 +41,19 @@ DEVICE_FOUND = 0x0012
 ADV_MONITOR_DEVICE_FOUND = 0x002F
 
 # Management commands
+MGMT_OP_DISCONNECT = 0x0014
 MGMT_OP_GET_CONNECTIONS = 0x0015
+MGMT_OP_PAIR_DEVICE = 0x0019
+MGMT_OP_UNPAIR_DEVICE = 0x001B
 MGMT_OP_START_DISCOVERY = 0x0023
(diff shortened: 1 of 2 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #564, opened 2026-06-23
feat: add mgmt load long term keys for bond restore

bdraco:
Adds load_long_term_keys to the bluez channel; it sends MGMT_OP_LOAD_LONG_TERM_KEYS (0x0013) with a uint16 key count followed by one 36 byte mgmt_ltk_info record per key, replacing the controller's LTK list so the kernel can re-encrypt links to already bonded peers without re-pairing. An empty list clears the keys.

Also adds the LongTermKey value type (a frozen dataclass for one mgmt_ltk_info: address, address type, key type, central flag, encryption size, ediv, 8 byte rand, 16 byte value) and the packing struct. The opcode and the 36 byte record layout were taken from the installed btsocket btmgmt_protocol; the struct size is asserted at 36 in the tests. The encoder validates each address and the rand/value lengths and treats a malformed key as a soft failure rather than raising into the command path.

This is the bond restore half of key persistence and reuses the same record layout the NEW_LONG_TERM_KEY event capture will decode in a follow up. It is a pure command sender like the pair/unpair/disconnect commands; nothing calls it yet, no behavior change, and there is no hot path or pxd change.

Tested the encoded count and the full 36 byte record field by field, the empty clear case, a failure status, and rejection of an invalid address or a wrong length rand/value with nothing sent; the new code is fully covered.

### DIFF
--- a/src/habluetooth/channels/bluez.py
+++ b/src/habluetooth/channels/bluez.py
@@ -4,6 +4,7 @@ import asyncio
 import logging
 from asyncio import timeout as asyncio_timeout
 from contextlib import asynccontextmanager
+from dataclasses import dataclass
 from functools import lru_cache
 from struct import Struct
 from typing import TYPE_CHECKING, Any, cast
@@ -41,6 +42,7 @@ DEVICE_FOUND = 0x0012
 ADV_MONITOR_DEVICE_FOUND = 0x002F
 
 # Management commands
+MGMT_OP_LOAD_LONG_TERM_KEYS = 0x0013
 MGMT_OP_DISCONNECT = 0x0014
 MGMT_OP_GET_CONNECTIONS = 0x0015
 MGMT_OP_PAIR_DEVICE = 0x0019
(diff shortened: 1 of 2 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Fixed. load_long_term_keys now wraps the LTK_INFO_PACK in try/except struct.error, so an out-of-range integer field (for example a corrupt key_type of 300) is logged and returns False rather than raising out of the command path, matching the soft-failure contract for the rest of the encoder. The explicit rand/value length checks stay since those silently pad/truncate rather than raise. Added a test with key_type 300 asserting False and nothing sent.

Left pair_device and unpair_device as is for now since they only pack address_type as a byte and the address type comes from a small known enum; if you want, I can give them the same guard in a follow up.

new code fully covered, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #565, opened 2026-06-23
feat(auto-scheduler): add last active window timing to diagnostics

bdraco:
Snapshot diagnostics could show active scans running and slots filled, but not whether the device's owning scanner actually ran a window on time; this fills that gap.

Per worker it adds last_window_at and last_window_duration. Per request address it adds last_active_window and last_active_window_source, so the scanning source can be compared against owner_source to tell whether the right scanner is active scanning the device, and how recently versus its scan_interval.

Motivated by home-assistant/core#174169, where scan response only Inkbird sensors update rarely behind multiple proxies; a single diagnostics snapshot could not confirm whether the owning scanner was active scanning on cadence.

### DIFF
--- a/src/habluetooth/auto_scheduler.pxd
+++ b/src/habluetooth/auto_scheduler.pxd
@@ -32,6 +32,8 @@ cdef class _ScannerWorker:
     cdef public object _task
     cdef public double _window_end
     cdef public double _sweep_last_completed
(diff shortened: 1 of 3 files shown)


bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

PR #566, opened 2026-06-24
feat: capture new long term key and auth failed mgmt events

bdraco:
Adds the receive side of pairing to the bluez channel. The management socket pushes NEW_LONG_TERM_KEY (0x000A) when bonding completes and AUTHENTICATION_FAILED (0x0011) when it does not; data_received now decodes both and routes them to a per peer handler. NEW_LONG_TERM_KEY is decoded into the existing LongTermKey record (so the client can persist it and feed it back via load_long_term_keys on reconnect) wrapped in a NewLongTermKey event with the kernel store hint; AUTHENTICATION_FAILED is decoded into an AuthenticationFailed event with the status.

Handlers are registered via MGMTBluetoothCtl.register_pairing_handler(adapter_idx, address, handler), which returns an unregister callback and normalizes the address. The handler dict is shared by reference with each protocol instance, like the scanners dict, so registrations survive a socket reconnect; the protocol looks up (controller_idx, address) and calls the handler. Event codes and the 36 byte mgmt_ltk_info layout came from the installed btsocket btmgmt_protocol; a truncated event is dropped with a debug log rather than mis-parsed.

This completes the bond persistence pair with the merged load_long_term_keys and unblocks the client capturing keys during pairing. The new event codes stay plain module constants since NEW_LTK and AUTH_FAILED are rare frames matched after the hot DEVICE_FOUND branch; the only pxd change is one cdef dict attribute on the protocol for the shared handler map. Nothing calls this yet, no behavior change.

Tested decode and dispatch of both events field by field, the no handler and truncated cases, and that register_pairing_handler normalizes the address and unregisters idempotently; the new code is fully covered. Verified on a cython build since this touches data_received and the protocol pxd.

### DIFF
--- a/src/habluetooth/channels/bluez.py
+++ b/src/habluetooth/channels/bluez.py
@@ -32,6 +32,8 @@ if TYPE_CHECKING:
 
     from ..base_scanner import BaseHaScanner
 
+    PairingHandler = Callable[["NewLongTermKey | AuthenticationFailed"], None]
+
 _LOGGER = logging.getLogger(__name__)
 _int = int
 _bytes = bytes
@@ -61,6 +63,8 @@ IO_CAPABILITY_NO_INPUT_NO_OUTPUT = 0x03
 # Management events
 MGMT_EV_CMD_COMPLETE = 0x0001
 MGMT_EV_CMD_STATUS = 0x0002
+MGMT_EV_NEW_LONG_TERM_KEY = 0x000A
+MGMT_EV_AUTH_FAILED = 0x0011
 
 # Discovery address-type bitmask: BR/EDR (0x01) | LE Public (0x02) | LE Random (0x04).
 # We only scan for LE devices, so we combine the two LE bits.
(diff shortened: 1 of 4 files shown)


src/habluetooth/channels/bluez.py
```diff
@@ -368,6 +408,44 @@ def data_received(self, data: _bytes) -> None:  # noqa: C901
...
+
+    def _handle_auth_failed(self, controller_idx: _int, param: _bytes) -> None:
+        """Decode an AUTHENTICATION_FAILED event and route it to its handler."""
+        if len(param) < _AUTH_FAILED_EV_SIZE:
+            _LOGGER.debug("hci%u: truncated AUTH_FAILED event", controller_idx)
+            return
+        address = bytes_mac_to_str(param[0:6])
+        if (handler := self._pairing_handlers.get((controller_idx, address))) is None:
+            return
+        handler(AuthenticationFailed(address, param[6], param[7]))
```
bdraco: need try / except here so a bad handler can't collapse the whole connection

src/habluetooth/channels/bluez.py
```diff
@@ -368,6 +408,44 @@ def data_received(self, data: _bytes) -> None:  # noqa: C901
...
+            _LOGGER.debug("hci%u: truncated NEW_LONG_TERM_KEY event", controller_idx)
+            return
+        store_hint = param[0]
+        addr, addr_type, key_type, central, enc_size, ediv, rand, value = (
+            LTK_INFO_UNPACK(param[1 : 1 + LTK_INFO_SIZE])
+        )
+        address = bytes_mac_to_str(addr)
+        if (handler := self._pairing_handlers.get((controller_idx, address))) is None:
+            return
+        handler(
```
bdraco: need try / except here so a bad handler can't collapse the whole connection

bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed your inline notes and the registration review.

Handler isolation (your two comments at the dispatch sites): both NEW_LTK and AUTH_FAILED now dispatch through _dispatch_pairing_event, which wraps the handler call in try/except and logs via _LOGGER.exception, so a raising handler can no longer collapse the read loop / the whole mgmt connection. Added a test that registers a handler which raises and asserts both event paths are isolated and logged.

register_pairing_handler: documented the last-wins semantics (a second registration for the same (adapter_idx, address) replaces the first, and the prior unregister becomes a no-op; handlers do not stack) and the expected canonical AA:BB:CC:DD:EE:FF address format (it is upper-cased to match the decoder output, but any other shape simply never matches). Added a last-wins test that also exercises the unregister identity guard.

Fully covered, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #567, opened 2026-06-24
feat(manager): warn when active scan is requested with a passive-only scanner

bdraco:
A passive-only scanner never runs an active window, so if it becomes the closest scanner for a device that only answers on its scan response, it will silently starve that device; there is no signal today that this is happening.

This warns once per source when an active scan is registered while a passive-only scanner exists, and when a scanner goes passive while active-scan requests already exist. The dedup keeps one passive proxy behind several devices from logging repeatedly, and the entry resets on unregister or when the scanner leaves passive mode.

Motivated by home-assistant/core#174169, where a 2nd ESPHome proxy left in passive mode took ownership of an Inkbird sensor and starved it with no warning.

### DIFF
--- a/src/habluetooth/auto_scheduler.py
+++ b/src/habluetooth/auto_scheduler.py
@@ -768,6 +768,11 @@ class AutoScanScheduler:
         self._on_demand_sweep_future: asyncio.Future[None] | None = None
         self._on_demand_sweep_end: float = 0.0
 
+    @property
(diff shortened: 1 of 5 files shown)


bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

PR #569, opened 2026-06-24
feat: wire mgmt pairing and bond restore into the gatt client

bdraco:
Makes HaMgmtClient.pair() and unpair() actually work, and restores bonds on connect, using the merged mgmt pairing codec.

pair() issues PAIR_DEVICE over the management socket and registers a pairing handler so the NEW_LONG_TERM_KEY pushed by the kernel during bonding is captured and persisted in a per adapter key store; a failed PAIR_DEVICE raises BleakError. unpair() issues UNPAIR_DEVICE and forgets the stored key. connect() now reloads a stored long term key via LOAD_LONG_TERM_KEYS before opening the L2CAP channel so the kernel re-encrypts a known peer without re-pairing, replacing the earlier relying on bonded keys placeholder.

The key store lives on HaScannerMgmt (per adapter, so bonds outlive the per connection client instances); the scanner hands the client the mgmt controller, the adapter index, and get/set/forget callbacks via MgmtClientData. All the new MgmtClientData fields are optional and additive, so a client built without them simply reports that pairing requires the management socket and skips bond restore.

This is the integration that lets the experimental DBus free client bond; it is still gated behind the experimental HaScannerMgmt and create_local_scanner, which Home Assistant does not select yet, so there is no behavior change. The passkey and confirmation interactive replies and honoring connect(pair=True) inline are follow ups.

Tested pair capturing and persisting the key, a non key event during pairing storing nothing, pairing failure raising, pairing without the mgmt socket raising, unpair removing the bond, connect restoring a stored key and loading nothing when unbonded, and the scanner key store; client_mgmt and scanner_mgmt both at 100 percent line and branch, full suite green.

### DIFF
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -405,7 +405,11 @@ class HaMgmtClient(BaseBleakClient):
     async def unpair(self) -> None:
         """Remove the bond over mgmt and forget the stored key."""
         mgmt, adapter_idx = self._require_mgmt()
-        await mgmt.unpair_device(adapter_idx, self.address, self._address_type)
+        if not await mgmt.unpair_device(adapter_idx, self.address, self._address_type):
+            # Don't drop the local key if the kernel bond is still there, or the
+            # store would silently desync from the controller.
+            msg = f"{self.address}: unpair failed"
+            raise BleakError(msg)
         if self._forget_long_term_keys is not None:
             self._forget_long_term_keys(self.address)
 
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -92,9 +92,13 @@ class MgmtClientData:
     # instance so reconnects can restore them).
     adapter_idx: int | None = None
     mgmt: MGMTBluetoothCtl | None = None
-    get_long_term_key: Callable[[str], LongTermKey | None] | None = None
-    set_long_term_key: Callable[[str, LongTermKey], None] | None = None
-    forget_long_term_key: Callable[[str], None] | None = None
+    # get_long_term_keys returns every bonded key for the adapter (LOAD_LONG_TERM_
+    # KEYS replaces the controller's whole list, so a restore must send them all);
+    # add_long_term_key stores one captured key; forget_long_term_keys drops all
+    # keys for a peer on unpair.
+    get_long_term_keys: Callable[[], list[LongTermKey]] | None = None
+    add_long_term_key: Callable[[LongTermKey], None] | None = None
+    forget_long_term_keys: Callable[[str], None] | None = None
 
 
 class HaMgmtClient(BaseBleakClient):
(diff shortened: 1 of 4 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- _restore_bond loads only the current peer's key, but LOAD_LONG_TERM_KEYS replaces the adapter's entire key list

bdraco:
Addressed the multi-bond design issue and the related items.

Important 1 (LOAD replaces the whole list): _restore_bond now loads every stored key for the adapter, not just the connecting peer's, so connecting one bonded device no longer wipes the others. The store interface changed to match: the scanner exposes get_long_term_keys (all keys), add_long_term_key, and forget_long_term_keys (all keys for a peer). Test now asserts all keys are loaded.

Suggestion 2 (multiple LTKs per peer): the scanner store is keyed by (address, central, ediv) instead of address, so a peer can hold more than one LTK without later events clobbering it; forget removes all keys for a peer. Tested.

Suggestion 1 (store_hint): _capture now only persists when store_hint is non-zero; a key the kernel asks not to keep is dropped. Tested.

HIGH silent failure (ignored load return): _restore_bond checks the bool and logs a warning when the bond reload fails, so a missing-bond reconnect is observable. Tested.

MEDIUM silent failure (captured key, no store): _capture now logs a warning when a key is captured but no store is wired, instead of silently dropping it. Tested.

adapter_idx parse: guarded with isdigit() so a malformed hci name yields None instead of raising.

On keying: bonds are per adapter (an LTK is tied to the local adapter identity it paired with, so it is not valid on another adapter); the store lives on HaScannerMgmt, one per adapter, which is correct. Disk persistence across restarts is a separate follow up; the store interface (get_all/add/forget callbacks) is persistence ready so a later store can back the same callbacks.

client_mgmt and scanner_mgmt both 100 percent line and branch, full suite green.

bluetoothbot (changes requested):
Blocking issues found.

- unpair forget is case-sensitive while stored keys are always upper-case

bdraco:
Addressed the case-normalization and the unpair return, and noted the two I left.

Important 1 / Copilot (case-sensitive forget): the scanner bond store now canonicalizes addresses to upper-case on both store and forget, so unpair removes the bond even if self.address is differently cased; this matches register_pairing_handler. Added a test that forgets with a lower-case address and asserts the upper-case stored key is removed.

HIGH / Copilot (unpair ignored the return): unpair() now checks the mgmt unpair_device bool and raises BleakError on failure, forgetting the local key only on success, so the store cannot desync from the controller; symmetric with pair(). Added a test that a failed unpair raises and leaves the stored key in place.

MEDIUM (restore failure proceeds): I kept this a warning rather than raising, because the L2CAP socket sets BT_SECURITY medium, so the kernel enforces encryption at link up; if the key did not load, the encrypted connection cannot be established and create_connection fails the connect anyway. So a failed restore does not silently produce an unencrypted link, the warning is the diagnostic and the transport is the enforcement.

adapter_idx suggestion: I can not reuse BaseHaScanner.adapter_idx here because the connector (and its client_data) is built before super().__init__ runs, so the property is not available yet; the local parse also adds the isdigit guard the base property lacks (it would raise on a malformed hci name).

client_mgmt and scanner_mgmt both 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #570, opened 2026-06-24
fix(manager): require durable staleness before a weaker scanner takes ownership

### LINKED ISSUE #568: Govee H5074 device data recorded incorrectly, H5075 works
## **Describe the bug**
I have Govee BLE H5075 and H5074 devices in my Home Assistant setup.  Starting around June 10th, the H5074 devices values have been sporadic and incorrect, causing my automations to behave improperly.

## **To Reproduce**
- Govee H5074 + H5075 in one room, ingested via govee_ble through one or more ESPHome bluetooth_proxy (active) nodes.
- Raise humidity quickly (shower). Watch the H5074 humidity entity oscillate/flatline while the H5075 tracks smoothly.
- (Optional, definitive) Capture raw adverts for the H5074 MAC with an independent ESP32 esp32_ble_tracker (active scan) and compare to the recorder — HA values will not match the air.

## **Additional context**
This problem was originally reported via https://github.com/home-assistant/core/issues/174507 and per feedback from @bdraco we are moving to this repo.

I performed an additional experiment based on the direction given by @bdraco and here is the data.  I do have more data, but I've tried to filter it down to what's absolutely necessary.  So if I've missed something, let me know and it's possible I have it and just didn't realize it was important.

Graph showing the 10minutes of the test and how the values changed over time:

Image

Data collected from my separate ESP32 capture device to show ground truth: [adverts.csv](https://github.com/user-attachments/files/29272745/adverts.csv)

Logs collected from habluetooth and homeassistant.components.bluetooth during the test: [ha_bt_debug_filtered.log](https://github.com/user-attachments/files/29272802/ha_bt_debug_filtered.log)

@bdraco also requested `along with govee_ble diagnostics for both the H5074 and the H5075` however I was unsure how to capture the BT diagnostics for just those devices, since I was only able to find a way to download the diagnostics of my BLE relays running on ESPHome.  Perhaps I need more guidance on how to do this?

In order to capture these deeper diagnostics, I asked Claude to help me try and get what our best guess of that data is, and this is what we captured:

H5074 data: [diag_H5074_A4C1384003B8_t03.json](https://github.com/user-attachments/files/29272891/diag_H5074_A4C1384003B8_t03.json)

H5075 data: [diag_H5075_A4C1382909C3_t03.json](https://github.com/user-attachments/files/29272900/diag_H5075_A4C1382909C3_t03.json)

### Claude's description of how these files were captured:

Exactly two HTTP-layer calls per file, against HA's diagnostics download API — not a govee_ble API (govee_ble has none). Here's the precise chain that produced
  both ..._t03.json files (the 3rd 60 s tick of your 18:27 run, captured at 2026-06-24T00:30:08Z):

  The API call

  Endpoint: GET /api/diagnostics/config_entry/{config_entry_id}
  config_entry_id: 01JKCQ1HMH5QB3HF4JJB18JBQK — the bluetooth integration entry for the plug-wyze-waterheater-down proxy.

  Because that view requires a signed URL, it takes two steps (in HA.diagnostics() / HA._sign() in run_shower_test.py):

  1. WebSocket ws://homeassistant.local:8123/api/websocket
    - authenticate with the bearer token
    - send {"type":"auth/sign_path","path":"/api/diagnostics/config_entry/01JKCQ1HMH5QB3HF4JJB18JBQK","expires":60}
    - receive back the same path with a ?authSig=<JWT> appended
  2. HTTP GET http://homeassistant.local:8123/api/diagnostics/config_entry/01JKCQ1HMH5QB3HF4JJB18JBQK?authSig=<JWT> (with Authorization: Bearer …) → the full
  ~4.7 MB JSON.

  That endpoint is the diagnostics integration's DownloadDiagnosticsView. For a bluetooth config entry it returns
  homeassistant.components.bluetooth.diagnostics.async_get_config_entry_diagnostics, whose data field is habluetooth's BluetoothManager.async_diagnostics() — the
  manager-wide dump. So every key you see (scanners[N].discovered_devices_and_advertisement_data, raw_advertisement_data, time_since_last_device_detection,
  connectable_history, all_history, advertisement_tracker, auto_scheduler) is emitted by habluetooth itself.

  What's mine vs. HA's

  - The single API response is global — one proxy's config entry, but the dump contains all 23 scanners' views of every device.
  - The per-MAC file is not a separate API call. extract_for_mac() in run_shower_test.py walks that one JSON locally and pulls out every node referencing
  A4:C1:38:40:03:B8 (H5074) or A4:C1:38:29:09:C3 (H5075). Both _t03 files came from the same 4.7 MB response (which is also saved whole as diag_FULL_last.json).

  Two NOTES:

  - These are bluetooth/habluetooth diagnostics, not govee_ble — which we suspect is the data @bdraco was looking for since the data is produced by habluetooth's manager.
  - Picking any of your 24 bluetooth entries yields the identical dump; I just pinned the waterheater-down entry. If bdraco would rather see it keyed to a different entry, it's a one-line change.

## **System information**

- **Home Assistant Core:** core-2026.6.4
- **OS:** Home Assistant OS 18.0 · Supervisor 2026.06.2 · Python 3.14.5 · generic-x86-64
- **Transport:** ~19 ESPHome `bluetooth_proxy` (active) nodes; ESPHome Device Builder add-on 2026.6.2. No local Bluetooth adapter — all BLE via proxies.

bdraco:
Scan-response-only sensors (Govee H5074, Inkbird IBS-TH2) behind many active proxies oscillate between fresh and stale readings. The closest proxy holds the current value but captures the scan response irregularly, so its receive time goes stale and a far weaker proxy carrying an older value steals ownership by the staleness check; RSSI then yanks it back, and the entity flip-flops.

Adverts carry no timestamp, so receive time is all we have. This keeps the stale handoff, but when the challenger is materially weaker than the owner it must wait until the owner is silent for stale_seconds times DURABLY_GONE_STALE_FACTOR (2.5, capped at the 900s fallback) before taking over. A comparable or stronger scanner still takes over at the normal stale window, so ordinary roaming is unchanged; only a device that moved into weak-only coverage hands off a little slower.

Fixes #568.

### DIFF
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -41,7 +41,7 @@ CONNECTABLE_FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS: Final = 195
 # the owner). A materially weaker scanner must wait until the owner has been
 # silent for stale_seconds * this factor (capped at
 # FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS) before it can take over. This
-# stops a far, weaker proxy from stealing ownership on a single missed
+# stops a far weaker proxy from stealing ownership on a single missed
 # interval and surfacing a stale capture for scan-response-only sensors
 # (e.g. Govee H5074) seen by many active proxies, while still letting a
 # device that genuinely moved into weak-only coverage hand off.
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -35,6 +35,18 @@ FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS: Final = 60 * 15
 # than BlueZ's.
 CONNECTABLE_FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS: Final = 195
 
+# When a different scanner's advertisement arrives after the owner's last
+# advertisement is stale, the owner is handed off immediately only if the
+# challenger is comparable-or-stronger (within ADV_RSSI_SWITCH_THRESHOLD of
+# the owner). A materially weaker scanner must wait until the owner has been
+# silent for stale_seconds * this factor (capped at
+# FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS) before it can take over. This
+# stops a far, weaker proxy from stealing ownership on a single missed
+# interval and surfacing a stale capture for scan-response-only sensors
+# (e.g. Govee H5074) seen by many active proxies, while still letting a
+# device that genuinely moved into weak-only coverage hand off.
+DURABLY_GONE_STALE_FACTOR: Final = 2.5
+
 
 # We must recover before we hit the 180s mark
 # where the device is removed from the stack
--- a/src/habluetooth/manager.pxd
+++ b/src/habluetooth/manager.pxd
@@ -9,6 +9,7 @@ cdef int NO_RSSI_VALUE
 cdef int ADV_RSSI_SWITCH_THRESHOLD
 cdef double TRACKER_BUFFERING_WOBBLE_SECONDS
 cdef double FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS
+cdef double _DURABLY_GONE_STALE_FACTOR
 cdef object FILTER_UUIDS
 cdef object AdvertisementData
 cdef object BLEDevice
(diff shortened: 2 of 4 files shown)


bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

PR #571, opened 2026-06-24
test(benchmark): cover cross-source advertisement arbitration

bdraco:
The existing inject benchmarks all use a single source, so they never reach the cross-scanner arbitration path; _should_keep_previous_adv short-circuits on the source check and _prefer_previous_adv_from_different_source is never called. A multi-proxy setup hits that path on every cross-source advertisement, so it is worth a baseline.

This adds two benchmarks: one where a strong owner keeps the device against a weaker second scanner (the keep decision, no ownership flip), and one where two equal scanners alternate with ever-later timestamps so ownership switches every injection (the switch decision). Together they cover both halves of the arbitration.

Landing this first gives a baseline; a follow-up that changes the arbitration (#570) can rebase on top so codspeed shows the real effect on this path. Relates to #568.

### DIFF
--- a/tests/test_benchmark_base_scanner.py
+++ b/tests/test_benchmark_base_scanner.py
@@ -169,6 +169,89 @@ async def test_inject_100_cross_source_keep_previous(
 
 @pytest.mark.usefixtures("enable_bluetooth")
 @pytest.mark.asyncio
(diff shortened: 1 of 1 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #572, opened 2026-06-24
feat: add bond key serialization and persistence hooks

bdraco:
Lets a consumer (Home Assistant) persist mgmt bonded keys to disk so they survive a restart without re-pairing. habluetooth is a library and does no disk I/O, so this provides the serialization plus the scanner hooks the consumer wires to its own Store, mirroring how the discovery cache is persisted today.

storage.py gains long_term_key_to_dict and long_term_key_from_dict plus the LongTermKeyDict TypedDict; the bytes fields serialize as hex, the same shape as the discovery helpers. HaScannerMgmt gains export_long_term_keys (snapshot for saving), restore_long_term_keys (seed on startup), and set_long_term_keys_changed_callback (fires when the bond store changes so the consumer can schedule a debounced save). LongTermKey, LongTermKeyDict, and the two functions are exported from the package.

Keys are per adapter (an LTK is tied to the local adapter identity it paired with), so the consumer persists keyed by adapter; nothing here writes to disk and nothing in Home Assistant selects this path yet, so there is no behavior change.

Added docs/bond_persistence.md with a worked Home Assistant wiring example (Store plus async_delay_save, keyed by adapter, using the change callback), matching the pattern HA already uses for remote scanner storage; the codespell hook now ignores hass so the example reads naturally.

Tested the to_dict/from_dict roundtrip and hex encoding, and the scanner export, restore, and change callback (fires on store and on a real forget, not a no-op, and can be cleared); scanner_mgmt and storage both 100 percent line and branch, full suite green.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,6 +41,10 @@ repos:
     rev: v2.4.2
     hooks:
       - id: codespell
+        args:
+          # "hass" is the Home Assistant instance variable, not a typo of "hash".
+          - -L
+          - hass
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v2.1.0
     hooks:
--- a/docs/bond_persistence.md
+++ b/docs/bond_persistence.md
@@ -1,0 +1,79 @@
+# Bond key persistence (experimental)
+
+> **Experimental.** This covers the DBus-free management/L2CAP path
+> (`HaScannerMgmt` + `HaMgmtClient`). The API is not stable and may change or be
+> renamed. Home Assistant does not select this path yet.
+
+When pairing over the BlueZ management socket, the kernel hands back a
+long-term key (LTK) per bond. `habluetooth` keeps these **in memory only** (it is
+a library and does no disk I/O), so without help they are lost on restart and
+the peer must be re-paired. To survive a restart, the consumer (Home Assistant)
+persists them and seeds them back on startup.
+
+Keys are **per adapter**: an LTK is tied to the local adapter identity it paired
+with, so it is only valid on that adapter. Persist keyed by adapter.
+
+## What `habluetooth` provides
+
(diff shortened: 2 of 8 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Made the store path symmetric with forget. _store_long_term_key now skips the change callback when the stored key is byte-for-byte identical to what is already under that (address, central, ediv) tuple, so a redundant store no longer schedules a save; a new or updated value still fires. Extended the change-callback test to cover identical store (no fire), same-tuple-new-value (fires), no-op forget (no fire), and real forget (fires).

scanner_mgmt 100 percent line and branch, full suite green.

bdraco:
Addressed the Copilot notes.

Callback isolation (scanner_mgmt:304): _notify_long_term_keys_changed now wraps the consumer callback in try/except and logs via _LOGGER.exception, so a faulty persistence hook can no longer propagate into a pair/unpair flow. Added a test where the callback raises and asserts the store call does not raise and the failure is logged.

TypedDict construction (storage:354): long_term_key_to_dict now builds LongTermKeyDict(...) instead of a plain dict literal, matching AdvertisementDataDict and BLEDeviceDict for better static typing.

Unchanged-store notify (scanner_mgmt:287): already handled in the previous commit; _store_long_term_key skips the callback when the key is byte-for-byte identical, so duplicate NEW_LONG_TERM_KEY events do not schedule a redundant save.

scanner_mgmt and storage both 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

bdraco:
Consolidated. Removed the private _all_long_term_keys, which duplicated export_long_term_keys, and pointed the client_data get_long_term_keys at the public export_long_term_keys, so there is a single source for the snapshot and no chance of drift.

scanner_mgmt 100 percent line and branch, full suite green.

PR #573, opened 2026-06-24
feat: add mgmt passkey and confirmation pairing replies

bdraco:
Adds the interactive-pairing codec for peers that do not do Just Works bonding. The kernel pushes USER_CONFIRMATION_REQUEST (numeric comparison) or USER_PASSKEY_REQUEST during pairing; data_received now decodes both and routes them to the same per peer pairing handler as NEW_LONG_TERM_KEY and AUTH_FAILED, and MGMTBluetoothCtl gains the four reply commands.

Events: UserConfirmationRequest (address, address_type, confirm_hint, value) and UserPasskeyRequest (address, address_type), dispatched through the existing isolated _dispatch_pairing_event so a raising handler cannot tear down the read loop. Commands: user_confirmation_reply(accept=True/False), user_passkey_reply(passkey), and user_passkey_negative_reply, sharing a small _user_reply helper that validates the address and reports success. Opcodes, the request event layouts, and the reply struct widths came from the installed btsocket btmgmt_protocol.

The new event codes stay plain module constants (rare frames, matched after the hot DEVICE_FOUND branch); no pxd change, the pairing handler map already exists. The client handler type was widened to the four event union; HaMgmtClient still only acts on NEW_LONG_TERM_KEY since it pairs NoInputNoOutput (Just Works), so confirm/passkey requests are not expected on its path. Nothing wires the new replies yet and Home Assistant does not select this path, so there is no behavior change.

Tested each reply command (confirm accept/reject, passkey reply with the uint32 passkey, passkey negative, failure status, invalid address) and both request events (decode and dispatch field by field, no handler, and truncated); the new code is fully covered, full suite green on a cython build.

### DIFF
--- a/src/habluetooth/channels/bluez.py
+++ b/src/habluetooth/channels/bluez.py
@@ -32,14 +32,6 @@ if TYPE_CHECKING:
 
     from ..base_scanner import BaseHaScanner
 
-    PairingHandler = Callable[
-        [
-            "NewLongTermKey | AuthenticationFailed"
-            " | UserConfirmationRequest | UserPasskeyRequest"
-        ],
-        None,
-    ]
-
 _LOGGER = logging.getLogger(__name__)
 _int = int
 _bytes = bytes
@@ -194,8 +186,11 @@ class UserConfirmationRequest:
 
     address: str
     address_type: int
-    confirm_hint: int  # non-zero: the value should be shown to the user
-    value: int  # the 6-digit numeric comparison value
+    # Per the mgmt API: when set, the value is not relevant and the user should
+    # just confirm the pairing (yes/no, no number to show); when 0, ``value``
+    # holds the number to display for comparison.
+    confirm_hint: int
+    value: int  # the 6-digit numeric comparison value (when confirm_hint is 0)
 
 
 @dataclass(slots=True, frozen=True)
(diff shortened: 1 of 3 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed both Copilot notes and your confirm_hint correction.

PairingHandler type (bluez:41): replaced the implicit string-literal concatenation with a proper PairingEvent type alias (NewLongTermKey | AuthenticationFailed | UserConfirmationRequest | UserPasskeyRequest) defined in a TYPE_CHECKING block after the event classes, and PairingHandler = Callable[[PairingEvent], None]; _dispatch_pairing_event uses PairingEvent too. No implicit concat, ruff and mypy clean.

_user_reply struct errors (bluez:996): the pack call is now wrapped in try/except struct.error, so an out-of-range address_type or passkey is a soft failure (logged, returns False) rather than escaping the command path, matching load_long_term_keys. Added a test with passkey 2**32 asserting False and nothing sent.

confirm_hint semantics (your note): you are right, the comment was inverted. Corrected it to the mgmt API meaning: when confirm_hint is set the value is not relevant and the user just confirms yes/no; when 0, value holds the number to display for comparison. Comment only, no runtime change.

New code fully covered, full suite green on a cython build.

bluetoothbot (approved):
No blocking issues found.

PR #575, opened 2026-06-24
feat: fall back to bleak when l2cap sockets are not permitted

bdraco:
The factory selected the mgmt scanner whenever mgmt discovery was available (can_discover, which reflects NET_ADMIN on the management socket), but the connect path opens a raw AF_BLUETOOTH L2CAP socket, whose creation needs a different capability (NET_RAW). A process that can discover but cannot open an L2CAP socket would get the mgmt scanner and then fail every connection with no fallback.

Adds can_use_l2cap() to the l2cap channel: a cheap probe that opens and closes an AF_BLUETOOTH SOCK_SEQPACKET socket and reports whether it succeeded. create_local_scanner now requires it, so a host that lacks L2CAP permission (or a kernel without Bluetooth) falls back to the bleak/DBus HaScanner and stays connectable.

The mgmt/L2CAP path is still experimental and not selected by Home Assistant, so no behavior change. Tested the probe both ways (socket opens, and PermissionError) and the factory falling back when L2CAP is unavailable; scanner_mgmt and l2cap both 100 percent line and branch, full suite green on a cython build.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## v6.21.0 (2026-06-24)
+
+### Features
+
+
+- Add mgmt passkey and confirmation pairing replies ([`84531ac`](https://github.com/Bluetooth-Devices/habluetooth/commit/84531acebac9378ec4447f559c9dd4d6676ab6f3))
+
+
 ## v6.20.0 (2026-06-24)
 
 ### Features
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -7,7 +7,7 @@
 project = "habluetooth"
 copyright = "2023, J. Nick Koston"
 author = "J. Nick Koston"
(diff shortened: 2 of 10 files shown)


bluetoothbot (approved):
No blocking issues found.

bdraco:
Addressed both.

Probe is now total (l2cap:113): the socket close moved into a try/finally, so a (very unlikely) close failure can no longer escape can_use_l2cap into create_local_scanner, and the descriptor cannot leak.

Silent downgrade (MEDIUM): the except OSError now debug logs the error before returning False, so a permission fallback is distinguishable from an unexpected failure (e.g. a kernel without Bluetooth) in production logs.

l2cap 100 percent line and branch, full suite green.

bluetoothbot (approved):
No blocking issues found.

PR #576, opened 2026-06-24
fix: stop the mgmt scanner watchdog from resurrecting a stopped scanner

bdraco:
The mgmt scanner watchdog schedules a restart on the start/stop lock; if a stop takes the lock first, the queued restart could still run afterwards and re-issue discovery on a scanner that was deliberately stopped.

This tracks whether discovery is meant to be running, skips the restart when it is not, and cancels any in-flight restart on teardown so it cannot resurrect discovery after the scanner is torn down.


bluetoothbot (approved):
No blocking issues found.

PR #577, opened 2026-06-24
chore: remove unused ATT_ERR_INVALID_HANDLE constant

bdraco:
ATT_ERR_INVALID_HANDLE is defined but referenced nowhere, neither in the codec nor in the tests, so it is dead.

The remaining ATT error codes are kept on purpose; some drive codec behavior, READ_NOT_PERMITTED is exercised by the tests as a representative error, and the encryption and authentication codes are retained for the upcoming on-demand security escalation.


bluetoothbot (approved):
No blocking issues found.

PR #578, opened 2026-06-24
feat: escalate l2cap security on demand to match bluetoothd

bdraco:
The raw L2CAP GATT client previously requested BT_SECURITY_MEDIUM up front on every connection, which forces encryption even for peers that do not need it. bluetoothd instead starts an ATT channel at the lowest level and escalates only when the peer rejects an operation.

This matches that behavior; the connection now starts at BT_SECURITY_LOW, and when an ATT request returns insufficient encryption (0x0F) or insufficient authentication (0x05) the codec asks the transport to raise the link security and re-issues the request once, exactly like bluetoothd's bt_att one-shot retry. 0x0F goes to MEDIUM, 0x05 steps up one level toward HIGH, and the level never downgrades. The codec stays transport agnostic, the error to level policy lives in the client, and the socket option lives in the transport.

Research basis: BlueZ src/shared/att.c change_security plus the op retry guard in handle_error_rsp.

### DIFF
--- a/src/habluetooth/channels/att.py
+++ b/src/habluetooth/channels/att.py
@@ -187,10 +187,16 @@ class ATTClient:
         self,
         send: Callable[[bytes], Awaitable[None]],
         on_disconnect: Callable[[Exception | None], None] | None = None,
+        escalate_security: Callable[[int], bool] | None = None,
     ) -> None:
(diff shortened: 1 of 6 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #579, opened 2026-06-24
refactor: dedupe mgmt command result checks in the bluez channel

bdraco:
The mgmt command wrappers (start/stop discovery, monitors, pair, unpair, disconnect, user replies, load keys) each repeated the same success check and the same no response or status logging tail, eight times over.

This extracts two small helpers, _command_succeeded and _result_status_text, and uses them throughout; the start discovery wrapper keeps its SUCCESS or BUSY special case but shares the status formatter. No behavior change.


bluetoothbot (approved):
No blocking issues found.

PR #580, opened 2026-06-24
fix(manager): keep a strong owner against a comparable scanner on stale

bdraco:
Follow-up to #570. On my own setup a stationary Inkbird/H5074 sitting right next to a strong proxy still flapped ownership ~8x in 9 minutes; every switch was a comparable proxy, so the data barely moved, but a device that never moves should not change owners.

Cause: the owner's normal ~10s reception gap (RF / scan-response jitter) barely beats the stale window, which is built from the aggregate advertising interval across all proxies and so underestimates a single owner's gap. A comparable proxy steals on the stale path, then the strong owner re-hears the device and reclaims on RSSI, twice per gap. The tracker cannot learn the real gap because the steal resets it.

Fix: when the owner's last advertisement was at least STRONG_OWNER_STALE_RSSI (-70 dBm), treat a brief silence as jitter; a merely-comparable challenger must wait for durable absence rather than steal on one missed interval. A weaker owner keeps the normal comparable-on-stale handoff, a materially stronger challenger still takes over immediately via RSSI, and #570's weak-steal block is unchanged.

Relates to home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -47,6 +47,16 @@ CONNECTABLE_FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS: Final = 195
 # device that genuinely moved into weak-only coverage hand off.
 DURABLY_GONE_STALE_FACTOR: Final = 2.5
 
+# An owner whose last advertisement was at least this strong is treated as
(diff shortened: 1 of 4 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #581, opened 2026-06-24
feat: bond on connect, encrypt bonded reconnects, and auto-confirm just works pairing

bdraco:
Three related improvements to the mgmt/L2CAP connect and pairing flow, from the latest review.

connect(pair=True) was ignored with only a debug log; it now drives the mgmt bond as part of connecting when the management socket is wired, so callers that pair then use get a real bond before MTU exchange and discovery. When mgmt is not wired it still connects and skips the bond.

A reconnect for a peer we already hold a long-term key for now opens the L2CAP socket at MEDIUM up front, so the kernel re-encrypts the link immediately like bluetoothd, instead of running MTU exchange and discovery in the clear until the first gated operation is rejected. On-demand escalation still backstops everything else.

pair() now auto-accepts a numeric-comparison confirmation, the only interaction a NoInputNoOutput controller can satisfy, so a peer that requests it does not stall the bond; and it fails fast on an authentication failure instead of waiting out the mgmt command timeout.

The stale module docstring (which claimed pairing was out of scope and connect ignored the pair flag) is corrected.

### DIFF
--- a/src/habluetooth/client_mgmt.py
+++ b/src/habluetooth/client_mgmt.py
@@ -444,7 +444,12 @@ class HaMgmtClient(BaseBleakClient):
         return self._mgmt, self._adapter_idx
 
     async def _pair_if_possible(self) -> None:
-        """Bond as part of connect when mgmt is wired; skip otherwise."""
+        """Bond as part of connect when needed and mgmt is wired."""
+        if self._has_stored_bond():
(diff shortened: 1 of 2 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- connect(pair=True) re-bonds an already-bonded peer, which the kernel rejects

bluetoothbot (approved):
No blocking issues found.

PR #582, opened 2026-06-24
fix: drop the notify handler when the stop_notify CCCD write fails

bdraco:
stop_notify removed the notify handler only after the CCCD-off write returned, so if that write failed the handler was left registered, still routing notifications for a subscription the caller had stopped.

The handler is now removed in a finally, mirroring start_notify's unwind, so a failed disable leaves no stale handler; the write error still surfaces to the caller.


bluetoothbot (approved):
No blocking issues found.

PR #583, opened 2026-06-24
docs: correct the stale att transport docstring and note the bond list limitation

bdraco:
Two documentation accuracy fixes from the review.

The att module docstring still said the L2CAP socket that feeds it would land in a later change; it landed, so it now points at L2CAPSocket.

load_long_term_keys now spells out that it replaces the controller's whole LTK list (the mgmt API has no add-one primitive), so if another mgmt agent such as bluetoothd manages the same adapter, loading our keys clears the bonds it loaded; the DBus-free path assumes it owns the adapter, so this is a documented coexistence limitation.


bluetoothbot (approved):
No blocking issues found.

PR #584, opened 2026-06-24
feat(manager): smooth per-source RSSI for advertisement owner selection

bdraco:
Follow-up to #570 and #580. On an all-stationary setup, devices behind several similar-distance proxies still flapped owner ~40 times in a window; nothing was moving, so it was pure RSSI noise. Multipath fades a fixed device's RSSI by 10 to 15 dB, so two proxies cross the 16 dB switch threshold on a single sample. This matters because, with the auto-scheduler, only the current owner actively scans a scan-response-only sensor, so each spurious flip hands the device to a proxy that has no fresh scan response yet.

This tracks an EWMA-smoothed RSSI per address and source and switches ownership on the smoothed value instead of the latest sample. A momentary spike barely moves the average so ownership stays put; a genuine move still crosses once the average shifts. The bucket is allocated only once an address is seen from more than one source, and it is evicted with the device and per source on unregister. The same smoothed value feeds the stale-path strong-owner and comparable checks for consistency.

On cost: a proxy-free host keeps the smoothed map empty, so every advert pays only an O(1) emptiness check and nothing else. Once any address on the host is seen from 2+ sources the map is non-empty, so single-proxy devices still pay that check plus a missed lookup and the source comparisons; they just never allocate a bucket.

Replaying the actual debug log, smoothing suppressed about 81% of the RSSI-path owner switches (855 down to 159); the busiest stationary devices dropped hardest, one going from 178 switches to 28, while genuinely moving devices still handed off. The single-source and proxy-free benchmarks stay flat. The two cross-source benchmarks regress about 20% and 12%, roughly 450 ns per advert: that path now maintains the EWMA and compares smoothed values on every advert from a device seen by 2+ proxies, which is the steady-state multi-proxy path. That cost is inherent to the feature and is acknowledged on CodSpeed; it is the deliberate trade for killing the per-window flapping.

Relates to home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -57,6 +57,16 @@ DURABLY_GONE_STALE_FACTOR: Final = 2.5
 # unaffected, and below typical close-proxy readings.
 STRONG_OWNER_STALE_RSSI: Final = -70
 
+# Smoothing factor (EWMA alpha) for the per-(address, source) advertisement
+# RSSI used to choose a device's owning scanner. Ownership switches on the
+# smoothed value rather than a single sample, so a stationary device whose
+# RSSI to a fixed proxy fades by multipath does not flap between proxies on
+# momentary spikes; a genuine move still crosses once the average shifts.
+# Smaller = steadier but slower to follow a real move. At 0.3 a single spike
+# moves the smoothed value ~30% of the gap, so even a 30 dB spike shifts it
+# under the 16 dB switch threshold.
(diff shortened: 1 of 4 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- Cross-source hot-path regression contradicts the "few ns" claim

bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

PR #585, opened 2026-06-24
feat(manager): add asymmetric rssi switch hysteresis to stop reclaim flapping

bdraco:
Follow up to #584, now merged. Once smoothing landed, the deployed build took the worst flapper from 178 owner switches to 10, but two heavily multipathed stationary devices still ping pong between two proxies; their smoothed RSSI genuinely straddles the 16 dB switch threshold, and each leg of the A to B to A switch only needs 16 dB.

This adds asymmetric hysteresis. A challenger pays an extra deadband (6 dB) only when it is reclaiming the ownership it was just demoted from, tracked per address in `_demoted_source`. A boundary device must swing its smoothed signal through a wider band to reclaim, so the oscillation stops; a genuine one way move is never the demoted source and still hands off at the plain 16 dB threshold, so real moves are not slowed.

The demotion is recorded inside the active RSSI branch of the predicate, so a stale path handoff records nothing; a weak owner that goes silent and is taken over on the stale path still reclaims at the plain threshold, not the deadband, so it is never left stuck on a weaker proxy. Recording there also means only a currently registered source is ever stored, since the predicate already checked it is in `_sources`. The state is evicted with the device and per source on unregister, mirroring `_smoothed_rssi`.

On the deployed log the two offenders' reclaim gaps cluster at 16 to 20 dB then jump to 22 dB and up, so this suppresses the boundary ping pong while leaving the genuine large smoothed crossings. Relates to #568 and home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -140,6 +140,20 @@ ADV_RSSI_SWITCH_THRESHOLD: Final = 16
 # selection that uses RSSI_SWITCH_THRESHOLD from
 # bleak_retry_connector
 
+# Extra margin, on top of ADV_RSSI_SWITCH_THRESHOLD, that a challenger must beat
+# the current owner by to RECLAIM ownership it was just demoted from, on the
+# active RSSI path (both scanners reporting). This is asymmetric hysteresis:
+# smoothing removes single-sample noise, but a heavily-multipathed stationary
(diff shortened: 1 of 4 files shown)


bluetoothbot (changes requested):
Blocking issues found.

- Demotion is recorded after stale-path handoffs too, charging the deadband to a returning owner

bluetoothbot (approved):
No blocking issues found.

PR #586, opened 2026-06-25
fix(manager): don't fold missing rssi into the smoothed average

bdraco:
On a bench with several co-located proxies, two stationary devices kept flapping owner; their smoothed RSSI to a single proxy was swinging 30 to 46 dB, which is impossible from geometry when the proxies sit right next to each other. The cause is on our side. When a proxy drops the RSSI on an advert, `service_info.rssi` is 0, and `rssi or NO_RSSI_VALUE` turns it into -127, which we then fold into the per-source EWMA. Because the average remembers it, one dropped reading drags the smoothed value down for several adverts (one -127 fold pulls a -17 reading to about -50, a second to about -73), manufacturing phantom weak readings that flap ownership. Smoothing made this worse than the old instantaneous path, which forgot the bad sample on the next packet.

Now a no-RSSI advert keeps the last good smoothed value and writes nothing, so it can never poison the average. A no-RSSI first cross-source sighting seeds only the existing owner and loses arbitration via NO_RSSI_VALUE rather than stealing ownership or priming the bucket with -127; the source's next real advert seeds it cleanly.

Found while validating the smoothing and hysteresis work from #584 and #585 against real bench logs. Relates to #568 and home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -880,8 +880,14 @@ class BluetoothManager:
         # just takes the miss-lookup below and skips the rest (its
         # new_smoothed is never read, since the predicate bails when the
         # bucket is None). Runs before the same-payload short-circuit below
-        # because RSSI changes even when the payload does not. A missing
-        # RSSI (0) is folded to NO_RSSI_VALUE, matching arbitration.
+        # because RSSI changes even when the payload does not.
+        #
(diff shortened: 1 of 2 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #589, opened 2026-07-01
fix(manager): damp owner reclaim across all recent owners, not just the last

bdraco:
Follow up to #585. On a bench with several co-located proxies, a scan-response-only Inkbird still showed occasional spikes to old data; the ownership was doing a 3-way (and wider) bounce that the deadband missed. #585 remembered only the single most recent demoted source, so in an A to B to C to A bounce, when A reclaims from C the stored source is B, not A; A is not seen as reclaiming, pays the plain 16 dB, and the switch fires, briefly surfacing a proxy with a stale scan response.

This remembers the set of sources currently in the demoted state instead of one. A challenger reclaiming any ownership it recently lost pays the deadband; a genuinely new or moving source, never a recent owner, still switches at 16 dB. The new owner leaves the set and the old owner joins it on each switch, so the current owner is never in its own set, the set is self bounded by the proxy count, and it is evicted with the device and per source on unregister.

In sustained N-way contention every proxy eventually enters the set, so switching among them all requires 16 plus the deadband; for co-located and stationary devices that stickiness is the intended behavior, and a device moving toward a new proxy is not slowed. A reclaim whose smoothed gap genuinely exceeds 16 plus the deadband still fires, so this only removes the sub-deadband bounce. Relates to #568 and home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/manager.pxd
+++ b/src/habluetooth/manager.pxd
@@ -13,6 +13,7 @@ cdef double _DURABLY_GONE_STALE_FACTOR
 cdef int _STRONG_OWNER_STALE_RSSI
 cdef double _RSSI_SMOOTHING_FACTOR
 cdef int _ADV_RSSI_SWITCH_DEADBAND
+cdef frozenset _EMPTY_DEMOTED
 cdef object FILTER_UUIDS
 cdef object AdvertisementData
 cdef object BLEDevice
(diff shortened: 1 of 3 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #590, opened 2026-07-01
fix(manager): keep a device on its owner while the owner is paused for a connection

bdraco:
A local BLE adapter cannot scan while it holds or establishes a connection; `connecting()` sets `scanning = not _connecting`. While paused it hears no advertisements for any device on that adapter, so the owner's last seen time ages. When it resumes, the device looks stale and a proxy steals ownership on the stale path, briefly surfacing the proxy's old scan response before the owner reclaims on RSSI. On a bench log this was a perfect 17 losses / 17 reclaims cycle on the local adapter, and on a scan response only Inkbird sensor it showed as thin vertical spikes to an old value. This is the residual of home-assistant/core#174169 that the RSSI smoothing and deadband work structurally could not reach.

Record when a scanner resumes scanning after a connection pause in `_last_scan_resume_time` (set on both connect success and failure, via the `connecting()` finally). Then keep a paused owner in arbitration rather than auto handing off on `scanning=False`, and skip the stale handoff while the owner is connecting or for one stale window after it resumes. The grace is bounded by the device's own stale interval, so a device that genuinely moved away still hands off afterward; the RSSI path is untouched, so a genuinely stronger challenger still wins; and #570 / #580 stale protections are unchanged. A never paused scanner has `_last_scan_resume_time = 0.0` and gets no grace.

Relates to #568 and home-assistant/core#174169.

### DIFF
--- a/src/habluetooth/base_scanner.pxd
+++ b/src/habluetooth/base_scanner.pxd
@@ -37,6 +37,7 @@ cdef class BaseHaScanner:
     cdef public unsigned int _connect_completed_total
     cdef public unsigned int _connect_failed_total
     cdef public double _last_connect_completed_time
+    cdef public double _last_scan_resume_time
 
     cpdef void _clear_connection_history(self) except *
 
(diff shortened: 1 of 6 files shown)


bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

Issue #591, opened 2026-07-03
Govee H5074 device data still recorded far less than device transmits


jrwagz:
##**Describe the bug**
This is a follow on to https://github.com/Bluetooth-Devices/habluetooth/issues/568 , which found many issues related to capturing data reliably from Govee H5074 devices.  Those fixes got deployed in HA Core 2026.7.0 and I reran the tests with the updated code.  The behavior is improved, but there is still room for improvement, and I'm coming here to understand what potential next steps are.

Specifically the bug is that my independent receiver captures far more samples from the Govee H5074 than Home Assistant does, and I think that can be improved.

Here is an image showing that the Govee H5074 device data is recorded at a much lower sample rate as compared to my independent receiver.

<img width="1680" height="1540" alt="Image" src="https://github.com/user-attachments/assets/f6bacec6-a110-407a-a05f-ad3456dad53b" />

What exactly is that independent receiver?  Here is the ESPHome YAML code for it:

[untether-bt-esp32-receiver.yaml](https://github.com/user-attachments/files/29648299/untether-bt-esp32-receiver.yaml)

##**To Reproduce**
Same steps as in #568 : 
- Govee H5074 + H5075 in one room, ingested via govee_ble through one or more ESPHome bluetooth_proxy (active) nodes.
- Raise humidity quickly (shower). Watch the H5074 humidity entity oscillate/flatline while the H5075 tracks smoothly.
- (Optional, definitive) Capture raw adverts for the H5074 MAC with an independent ESP32 esp32_ble_tracker (active scan) and compare to the recorder — HA values will not match the air.

##**Additional context**
Captured diagnostic information:

[diag_FULL_last.json](https://github.com/user-attachments/files/29648334/diag_FULL_last.json)

[ha_bt_debug_filtered.log](https://github.com/user-attachments/files/29648338/ha_bt_debug_filtered.log)

jrwagz:
While the version of habluetooth in HA core 2026.7.0 was an improvement, the version in 2026.7.1 has made this worse. @bdraco 

Here are my test results on the latest release, I ran two tests back to back just to verify that the first bad result wasn't a one-off fluke.

For comparison, I ran this test on 2026.7.0 about 4 times in total and all 4 of those look very similar to the results posted above, whereas 2026.7.1 is performing worse, as shown in the results below.

## Test 1

[...]

PR #593, opened 2026-07-04
fix(manager): damp stale handoffs per device class and give rescues an accept grace

bdraco:
Field data from a dev instance running 6.26.3 showed the immediate stale handoffs flapping ownership: in two minutes there were 15 passive stale steals, some exceeding the stale threshold by as little as 11ms, all by materially weaker scanners (for example -97.7 stealing from -79.8) that happened to catch the one advertisement the owner missed; the strong owner then re-heard the device within seconds and reclaimed on the RSSI path, cycling every ~10s per device. The stale path ignores the reclaim deadband, so the hysteresis never damps this. Comparable band challengers have the mirrored problem; they cannot be reclaimed on the RSSI path at all, so a stale steal ping-pongs on alternating missed intervals, the stationary device flap #568 and #580 fixed.

This removes the two immediate stale handoff fast paths #592 added (passive newer capture wins, ACTIVE challenger wins) and restructures the stale decision per device class. An owner silent past the durably gone threshold loses to any challenger, unconditionally. Short of that, a passive device keeps the pre #591 roaming rule, a comparable or stronger challenger of a weak owner switches instantly and anything else waits for durably gone; payloads are identical across scanners so waiting costs nothing. An active need device never takes the roaming shortcut on the all history decision, even a comparable challenger may be an AUTO scanner sitting in its passive phase whose capture carries no fresh scan response (the issue #568 stale capture class); every handoff before durably gone goes through the rescue flow from #592, so it lands on a capture from an address a recent active window covered. The connectable re check keeps the ordinary roaming rule for every device class; connection routing cares about liveness, not scan response freshness, so the connectable history still repoints away from a silent scanner at the stale window.

Rescue scans only ever start on AUTO scanners. The owner side is always served (an AUTO owner gets its window, an ACTIVE owner counts as covered); a challenger that is already actively scanning needs no window, its side just records coverage. Accept times uniformly sit RESCUE_SCAN_ACCEPT_SECONDS after coverage, single sourced inside the recording helper, and the handoff additionally requires one full advertising interval since the trigger: the 5s grace alone guarantees nothing for a slow advertiser that cannot re-transmit within it, so the interval requirement is what makes the owner re-hear guarantee real for every cadence and removes the race where owner and challenger both hear the next advertisement. The retry only restarts an episode whose window never materialized; a recorded accept means the episode is just waiting out the gates, and in practice the rescue accelerates handoffs for advertisers faster than roughly the retry interval while the durably gone backstop governs slower ones. A rescue episode's lifetime is bounded by the active scan need: unregistering it clears the episode, so a device that never goes stale again cannot keep the scheduler's no episode fast path disabled forever.

The smoothing from #584, the reclaim deadband from #585 and #589, and the strong owner damping from #580 are untouched and continue to apply to all device classes.

Follow up to #592, relates to #591.

### DIFF
--- a/src/habluetooth/auto_scheduler.pxd
+++ b/src/habluetooth/auto_scheduler.pxd
@@ -183,7 +183,7 @@ cdef class AutoScanScheduler:
     cpdef void _rescue_challenger_side(
         self,
         str address,
-        str challenger_source,
+        object challenger,
         set requests,
         object loop,
         double now,
--- a/src/habluetooth/auto_scheduler.pxd
+++ b/src/habluetooth/auto_scheduler.pxd
@@ -147,14 +147,15 @@ cdef class AutoScanScheduler:
         self, str address, str challenger_source, str owner_source
     )
 
-    cpdef void _record_rescue_accept(self, str address, double accept_after)
+    @cython.locals(accept_after=double)
+    cpdef void _record_rescue_accept(self, str address, double coverage_time)
 
     cpdef void _record_window_dispatch(
         self,
         str address,
         double dispatch_now,
         str source,
-        double accept_after,
+        double coverage_time,
     )
 
     @cython.locals(address=str)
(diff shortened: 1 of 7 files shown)


bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

bluetoothbot (approved):
No blocking issues found.

PR #594, opened 2026-07-04
fix(manager): require one and a half stale windows before a roam handoff

bdraco:
Field data from a fully stationary network on 6.26.4 showed the remaining ownership churn is driven by a per scanner reception lottery, not device movement. Scanners duty cycle their scan windows, so each proxy only catches a fraction of a device's advertisements; "the owner missed an interval" is a per scanner event, and on the first miss the roaming rule parked ownership on a weak scanner that the strong owner then reclaimed on the RSSI path. Most of the observed RSSI path switches were these recoveries; the switch logs consistently show a strong winner reclaiming from a much weaker owner that had just acquired the device on the stale path.

The roaming rule (a comparable or stronger challenger of a weak owner) now requires the owner to have been silent for STALE_ROAM_FACTOR (1.5) stale windows instead of one, so the owner must effectively miss two reception opportunities before a roam. A genuinely departed device still roams at one and a half intervals, well before the durably gone backstop, which stays at DURABLY_GONE_STALE_FACTOR (2.5) unchanged. The same gate applies to the connectable re check so connection routing still repoints away from a silent scanner ahead of durably gone. Active need devices are untouched; they keep the rescue flow and never take the roaming shortcut.

Note for the next core bump: the bluetooth manager tests in Home Assistant that assert a stale handoff at one stale window (test_switching_adapters_based_on_stale variants) will need their switch timings moved past the 1.5x gate, mirroring the test updates here.

Follow up to #593, relates to #591.

### DIFF
--- a/src/habluetooth/const.py
+++ b/src/habluetooth/const.py
@@ -47,6 +47,20 @@ CONNECTABLE_FALLBACK_MAXIMUM_STALE_ADVERTISEMENT_SECONDS: Final = 195
 # device that genuinely moved into weak-only coverage hand off.
 DURABLY_GONE_STALE_FACTOR: Final = 2.5
 
+# How many stale windows must elapse before the roaming rule (a
+# comparable-or-stronger challenger of a weak owner) may take a stale
+# device. Scanners duty-cycle their scan windows, so "the owner missed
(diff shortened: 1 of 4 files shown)


bluetoothbot (approved):
No blocking issues found.

PR #607, opened 2026-08-18
fix: notify allocation callbacks when a scanner is unregistered

bdraco:
Unregistering a scanner popped its entry from the allocations map silently, so websocket subscribers kept rendering the removed scanner's last known allocated addresses forever; connectable scanners also had no allocations entry at all until the first firmware push, which made a reconnected proxy show stale or missing state. Part of the confusing diagnostics in home-assistant/core#176516.

Unregister now drops the entry and dispatches a zeroed allocations snapshot to subscribers, so a reentrant read from inside a callback already sees it gone, and registration seeds a zeroed entry for every scanner, visible via the pull API immediately with the first callback firing when slot information is actually reported, preserving a push that legitimately arrived before registration since bleak-esphome subscribes to slot updates before the scanner is registered. A slots value of zero remains the established sentinel for no slot information reported, matching how non connectable scanners have always been seeded and how consumers filter on slots greater than zero before reasoning about exhaustion.

Registering a scanner whose source is already registered by a live scanner is always a caller bug (the source map is last writer wins and the first unregister removes the live scanner), so registration now logs an error naming both scanners rather than corrupting silently; a normal re register after unregister stays quiet.

### DIFF
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1925,7 +1925,10 @@ class BluetoothManager:
         dispatched so subscribers stop rendering its stale addresses;
         ``HaScannerRegistrationEvent.REMOVED`` on
         ``async_register_scanner_registration_callback`` disambiguates
-        removal from an empty but present scanner.
+        removal from an empty but present scanner. Registration seeds a
+        zeroed entry without dispatching: it is visible immediately via
+        ``async_current_allocations`` and the first callback fires when
+        slot information is actually reported.
         """
         self._allocations_callbacks.setdefault(source, set()).add(callback)
         return partial(
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -121,6 +121,11 @@ def _dispatch_bleak_callback(
         _LOGGER.exception("Error in callback: %s", bleak_callback.callback)
 
 
+def _zeroed_allocations(source: str) -> HaBluetoothSlotAllocations:
+    """Return an empty allocations snapshot for a source."""
+    return HaBluetoothSlotAllocations(source=source, slots=0, free=0, allocated=[])
+
+
 class BleakCallback:
     """Bleak callback."""
 
@@ -1627,7 +1632,7 @@ class BluetoothManager:
         for address in emptied_demoted:
             del self._demoted_sources[address]
         self._adapter_sources.pop(scanner.adapter, None)
-        self._allocations.pop(scanner.source, None)
+        self._async_clear_allocations(source)
         if connection_slots:
             self.slot_manager.remove_adapter(scanner.adapter)
         if (idx := scanner.adapter_idx) is not None:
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Suggestions 2 and 3 applied in the latest push; the unregister test now also pins the global source=None subscriber that Home Assistant's websocket uses by default, and the private method test is replaced by a register, unregister, re-register round trip through the public API that asserts the re-seed. Since every scanner is now seeded at registration the pop guard in the clear helper was unreachable, so it is gone rather than tested.

On suggestion 1, I looked at whether the preserve on register guard could keep a stale post-unregister push and I do not think that sequence is reachable, so I left it without a grace period heuristic. bleak-esphome subscribes to slot pushes on the APIClient before registering the scanner, and both the subscription and the scanner registration die with the same API connection; on ESP reconnect a fresh ESPHomeBluetoothDevice subscribes and pushes before the new scanner registers. So a push landing between unregister and re-register can only come from the new session's subscription, which is exactly the pre-registration push the guard exists to keep, never a leftover from the previous session. A leftover would require the old subscription to fire after its own connection tore it down, which aioesphomeapi does not do. If a different glue ever pushed out of band the same guard would still be right for it since a push always carries current proxy state.

The CodSpeed regression on test_inject_100_cross_source_keep_previous looks like cross environment variance rather than a real change; the report itself flags different runtime environments and shows equal magnitude improvements on sibling inject benchmarks, and this diff touches only scanner registration and allocation callback dispatch, nothing on the advertisement inject path those benchmarks measure.

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Restored the tolerant pop in the clear helper so a duplicate source unregister can never abort teardown midway, and added a test that registers two scanners under one source, unregisters both, and asserts both REMOVED events fire and the allocations end empty.

bdraco:
@bluetoothbot review

bdraco:
Registering a scanner with a source that is already registered by a live scanner is always a caller bug (the source map is last writer wins and the first unregister removes the live scanner), so registration now logs an error naming both scanners; it does not raise so a scanner setup path is not aborted in production for it. A normal re register after unregister stays quiet, both cases are pinned in the tests.

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Both notes handled. The zeroed seed keeps the long standing semantics, slots zero has always meant no slot information reported (non connectable scanners were seeded that way before this PR, and every exhaustion consumer filters on slots greater than zero), so I documented the sentinel on the helper and on async_current_allocations rather than special casing which scanners get seeded; seeding on get_allocations would drop exactly the proxies the row is for, since it returns None until the first push. The PR description now covers the duplicate source error log. On the message wording, the explicit source stays because scanner names are not guaranteed to embed it.

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Documented the seed as pull only on async_register_allocation_callback, the entry is visible via async_current_allocations at registration and the first callback fires when slot information is actually reported, and fixed the description's ordering wording, the code pops then dispatches which is the better order for reentrant reads.

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

PR #609, opened 2026-08-21
fix: disconnect clients when their scanner is unregistered

### LINKED ISSUE #608: connections are not disconnected when their scanner is unregistered
### What happens

Unregistering a scanner clears everything habluetooth knows about it — advertisement
tracking, smoothed RSSI, demoted sources, the allocations entry, the slot-manager
adapter — but it never touches the connections that were made *through* that scanner.

For a local adapter those are BlueZ D-Bus links. BlueZ has no idea Home Assistant
considers the adapter gone, the kernel keeps the link up, and the integration holding
the `BleakClient` is never told anything. So the device keeps working through an
adapter that has been removed.

### How I ran into it

Home Assistant, one Raspberry Pi, three transports available: a TP-Link UB500 USB
dongle (`hci0`), the Pi's built-in `bcm43438` (`hci1`), and an ESPHome Bluetooth
proxy. I was benchmarking which of them each of my devices works best through, so I
disabled and enabled the `bluetooth` config entries one at a time.

A Truma iNet X panel (a connected GATT device with notifications) kept reporting
fresh, *changing* values in every configuration — including one where I had disabled
**every** transport:

```
bluetooth/subscribe_scanner_details  ->  []          # no scanners at all

sensor.truma_..._room_temperature
  14:12:22  32.9
  14:12:38  32.9
  14:12:53  32.2      <- real notification frames, ~15 s apart
  14:13:09  32.9
  14:13:24  32.8
```

Nine updates over two minutes with zero registered scanners. The link was still on
`hci0` from before I disabled it.

`bluetooth/subscribe_connection_allocations` shows the other half of the problem —
while proxy-only was active, the proxy reported:

```json
{"source":"B4:BF:E9:62:27:02","slots":3,"free":2,"allocated":["C4:D3:6A:8C:B5:38"]}
```

The Truma panel (`50:98:93:FF:B4:D1`) appears nowhere, because the source that owned
it was popped from `_allocations` on unregister — while the connection itself was
still up and still occupying a real slot on `hci0`.

### Why it matters

1. **The link outlives the adapter.** A user who disables an adapter, or switches
   between adapters, sees a device that still works and reasonably concludes the new
   transport is carrying it. It isn't. This makes adapter A/B testing silently
   meaningless, which is how I found it.

2. **Slot accounting desyncs.** `_async_clear_allocations(source)` and
   `slot_manager.remove_adapter(scanner.adapter)` drop the record of a connection that
   still exists. Re-register the adapter and it starts at `free == slots` with that
   link still up, so the budget can be overcommitted by however many connections
   survived. The resulting connect failures look like device problems.

This is consistent with what `CLAUDE.md` already notes under *Gotchas*: there is no
"currently-connected" counter, only `_connect_in_progress`. The gap is that unregister
assumes forgetting is equivalent to tearing down.

### Where

`manager.py`, `_async_unregister_scanner_internal`. Everything else about the scanner
is cleaned up there; connections are the one thing that is only forgotten.

Note `HaBleakClientWrapper` already records `_connected_scanner` for every live client,
and the manager already emits `HaScannerRegistrationEvent.REMOVED` — the information
needed to do this is present, nothing consumes it for teardown.

### Fix

Verified against real BlueZ on the setup above: with the patch, disabling the
adapter's config entry drops the kernel link within ~3 s (`btmgmt con` and
`/sys/kernel/debug/bluetooth/l2cap` both go empty), the integration reconnects
cleanly when the adapter is re-enabled, and nothing raises.

Tracks connected clients per source in a `WeakSet` and disconnects them when the
scanner is unregistered. Weak so a client dropped without an explicit disconnect
cannot leak; an explicitly disconnected client unregisters itself.

### Versions

- habluetooth `6.26.7` (reproduced on `main` @ `8d3e16c`)
- Home Assistant with local BlueZ adapters + an ESPHome proxy
- Python 3.14, Linux

rpodgorny:
Fixes #608.

`_async_unregister_scanner_internal` tears down every piece of bookkeeping about a
scanner — advertisement tracking, smoothed RSSI, demoted sources, allocations, the
slot-manager adapter — but never touches the connections made through it.

For a local adapter those are BlueZ D-Bus links. BlueZ has no idea Home Assistant
considers the adapter gone, the kernel keeps the link up, and the integration holding
the client is never told. Two consequences:

- the device keeps working through an adapter the user removed, so switching or
  disabling adapters appears to work when nothing is actually carrying the traffic;
- the allocation bookkeeping that was just cleared has forgotten a connection that
  still occupies a real slot. On re-register the adapter starts at `free == slots`
  with that link still up, so the slot budget can be overcommitted.

Observed on a Home Assistant install with two local adapters and an ESPHome proxy: a
connected GATT device kept delivering notifications with `subscribe_scanner_details`
returning `[]` — every transport disabled — because its link was still open on the
adapter that had been unregistered minutes earlier.

### What this does

- `BaseHaScanner` gains `_clients: WeakSet[HaBleakClientWrapper]`, populated by
  `HaBleakClientWrapper.connect()` once a connection is established.
- `_async_unregister_scanner_internal` disconnects everything still connected
  through that scanner, bounded by a timeout and skipping clients that have
  reconnected through another scanner.
- The set is weak, so a client dropped without an explicit disconnect cannot leak.
- `HaBleakClientWrapper.disconnect()` unregisters itself, so an already-disconnected
  client is not disconnected twice.
- `base_scanner.pxd` and `manager.pxd` updated for the new attributes.

Caller visible changes: ``connect()`` raises ``BleakError`` when the scanner
was unregistered while the connect was in flight (after tearing the link down
inline); ``disconnect()`` clears ``_connected_scanner`` /
``_connected_device`` on a clean return (kept only while a raise leaves the
link up); and ``set_connection_params()`` on a disconnected local client is a
debug logged no-op instead of a kernel conn param write or a misleading
"upgrade the backend" warning (backends that implement it keep their own
not connected handling). When a forced teardown gives a client up, the
consumer's disconnected callback fires so integrations schedule their
reconnect.

### Tests

New tests in `tests/test_wrappers.py` (plus a parametrized in flight matrix):

- `test_unregister_scanner_disconnects_only_its_clients`
- `test_disconnected_client_is_not_disconnected_again`
- `test_client_tracking_does_not_keep_client_alive`
- `test_disconnect_error_on_unregister_is_logged_not_raised`
- `test_client_reconnected_elsewhere_is_not_disconnected`
- `test_reconnect_untracks_from_previous_scanner`
- `test_disconnect_timeout_on_unregister_is_logged`
- `test_stop_cancels_pending_disconnects`
- `test_manager_backend_timeout_is_not_misreported`
- `test_give_up_detaches_the_disconnected_callback`
- `test_give_up_survives_a_backend_that_refuses_to_detach`
- `test_failed_disconnect_with_link_down_untracks`
- `test_failed_disconnect_keeps_client_tracked`
- `test_connect_in_flight_when_scanner_unregisters`
- `test_dropped_link_is_not_disconnected_on_unregister`
- `test_track_moves_client_between_scanners`
- `test_no_disconnects_scheduled_after_stop`
- `test_background_task_exception_is_logged`
- `test_disconnect_untracks_even_if_backend_still_reports_connected`
- `test_base_exception_from_disconnect_gives_up_client`
- `test_set_connection_params_disconnected_is_quiet`
- `test_disconnect_without_backend_untracks_stale_pointers`
- `test_disconnect_without_connected_scanner`
- `test_disconnect_keeps_other_clients_tracked`
### Verified against real BlueZ

Not just unit tests. Home Assistant 2026.8.1 on a Raspberry Pi (aarch64), real
TP-Link UB500 adapter, real device (a Truma iNet X panel holding a GATT
connection with notifications). Patched build installed, adapter's config entry
disabled, kernel link state polled once a second:

```
17:14:00  con=1 l2cap=2
17:14:02  con=1 l2cap=2      <- config entry disabled at 17:14:01
17:14:03  con=1 l2cap=2
17:14:04  con=0 l2cap=0      <- link gone
17:14:15  con=0 l2cap=0
```

(`con` = `btmgmt --index 0 con`, `l2cap` = matching lines in
`/sys/kernel/debug/bluetooth/l2cap`.)

On the unpatched build the same action left the link up indefinitely — measured
at two minutes, still delivering notifications, with
`bluetooth/subscribe_scanner_details` returning `[]`.

Re-enabling the entry reconnected cleanly 11 s later. The integration holding the
client handled the forced disconnect without complaint — it logged
`BLE link closed cleanly`, correctly reported `no route to the panel right now`
while no transport existed, and backed off (15 s → 30 s) until the adapter
returned. **Zero tracebacks in the Home Assistant log across the whole test.**

### Points worth a maintainer's opinion

1. **Remote scanners.** This also disconnects clients of an ESPHome proxy when its
   scanner unregisters. Only the local-adapter path is covered by the real-world
   test above. I believe the proxy path is correct — connections through a proxy that
   has gone away are dead anyway, and this makes HA notice immediately instead of
   waiting for a timeout — but it is a behaviour change for that path too, so worth a
   look.

2. **Reload churn.** HA unregisters on config-entry unload, which includes *reload*.
   Reloading a bluetooth entry now drops its connections; integrations reconnect. That
   seems right ("the adapter is going away") but it is more churn than before.

3. **Alternative shape.** The manager already emits `HaScannerRegistrationEvent.REMOVED`
   and `HaBleakClientWrapper` already knows its `_connected_scanner`, so this could
   instead be pushed to consumers via the existing event rather than done centrally.
   I went central because every consumer would otherwise have to reimplement it, but
   happy to reshape.

### DIFF
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -64,7 +64,7 @@ from .usage import install_multiple_bleak_catcher, uninstall_multiple_bleak_catc
 from .util import async_reset_adapter, coalesce_concurrent_future
 
 if TYPE_CHECKING:
-    from collections.abc import Callable, Iterable
+    from collections.abc import Callable, Coroutine, Iterable
 
     from bleak.backends.device import BLEDevice
     from bleak.backends.scanner import AdvertisementData, AdvertisementDataCallback
@@ -1667,17 +1667,21 @@ class BluetoothManager:
         self._auto_scheduler.remove_scanner(scanner)
         self._async_on_scanner_registration(scanner, HaScannerRegistrationEvent.REMOVED)
 
+    def async_add_background_task(self, coro: Coroutine[Any, Any, None]) -> None:
+        """Run a coroutine as a background task that is cancelled on stop."""
+        if TYPE_CHECKING:
+            assert self._loop is not None
+        task = self._loop.create_task(coro)
+        self._background_tasks.add(task)
+        task.add_done_callback(self._background_tasks.discard)
+
     def _async_disconnect_clients(self, scanner: BaseHaScanner) -> None:
         """Disconnect the clients still connected through a removed scanner."""
         clients = list(scanner._clients)
         scanner._clients.clear()
         if not clients:
             return
-        if TYPE_CHECKING:
-            assert self._loop is not None
-        task = self._loop.create_task(self._async_disconnect_all(clients, scanner))
-        self._background_tasks.add(task)
-        task.add_done_callback(self._background_tasks.discard)
+        self.async_add_background_task(self._async_disconnect_all(clients, scanner))
 
     async def _async_disconnect_all(
         self, clients: list[HaBleakClientWrapper], scanner: BaseHaScanner
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1669,11 +1669,13 @@ class BluetoothManager:
         # just cleared has forgotten them.
         self._async_disconnect_clients(scanner)
 
-    def async_add_background_task(self, coro: Coroutine[Any, Any, None]) -> None:
+    def async_add_background_task(
+        self, coro: Coroutine[Any, Any, None], name: str
+    ) -> None:
         """Run a coroutine as a background task that is cancelled on stop."""
         if TYPE_CHECKING:
             assert self._loop is not None
-        task = self._loop.create_task(coro)
+        task = self._loop.create_task(coro, name=name)
         self._background_tasks.add(task)
         task.add_done_callback(self._on_background_task_done)
 
@@ -1689,7 +1691,10 @@ class BluetoothManager:
             return
         clients = list(scanner._clients)
         scanner._clients.clear()
-        self.async_add_background_task(self._async_disconnect_all(clients, scanner))
+        self.async_add_background_task(
+            self._async_disconnect_all(clients, scanner),
+            f"disconnect clients of {scanner.source}",
+        )
 
     async def _async_disconnect_all(
         self, clients: list[HaBleakClientWrapper], scanner: BaseHaScanner
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -475,6 +475,14 @@ class BluetoothManager:
             self._cancel_unavailable_tracking.cancel()
             self._cancel_unavailable_tracking = None
         self._auto_scheduler.stop()
+        if self._background_tasks:
+            # The cancel is fire and forget; log here so an unregister that
+            # raced shutdown leaves a trace even if the task never started.
+            _LOGGER.debug(
+                "Cancelling %d background task(s) at stop: %s",
+                len(self._background_tasks),
+                [task.get_name() for task in self._background_tasks],
+            )
         for task in list(self._background_tasks):
             task.cancel()
         uninstall_multiple_bleak_catcher()
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1677,7 +1677,13 @@ class BluetoothManager:
             assert self._loop is not None
         task = self._loop.create_task(coro)
         self._background_tasks.add(task)
-        task.add_done_callback(self._background_tasks.discard)
+        task.add_done_callback(self._on_background_task_done)
+
+    def _on_background_task_done(self, task: asyncio.Task[None]) -> None:
+        """Drop a finished background task, logging an escaped exception."""
+        self._background_tasks.discard(task)
+        if not task.cancelled() and (exc := task.exception()) is not None:
+            _LOGGER.error("Background task failed", exc_info=exc)
 
     def _async_disconnect_clients(self, scanner: BaseHaScanner) -> None:
         """Disconnect the clients still connected through a removed scanner."""
@@ -1714,6 +1720,14 @@ class BluetoothManager:
                 return
             async with asyncio.timeout(CLIENT_DISCONNECT_TIMEOUT):
                 await client.disconnect()
+        except TimeoutError:
+            # The expected shape for a proxy that went away; no traceback.
+            device = client._connected_device
+            _LOGGER.warning(
+                "Timed out disconnecting client %s from removed scanner %s",
+                device.address if device else "unknown",
+                scanner.source,
+            )
         except Exception:  # pylint: disable=broad-except
             device = client._connected_device
             _LOGGER.exception(
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -10,7 +10,6 @@ import platform
 from dataclasses import asdict
 from functools import partial
 from typing import TYPE_CHECKING, Any
-from weakref import WeakSet
 
 from bleak_retry_connector import (
     NO_RSSI_VALUE,
@@ -162,11 +161,11 @@ class BluetoothManager:
         "_allocations",
         "_allocations_callbacks",
         "_auto_scheduler",
+        "_background_tasks",
         "_bleak_callbacks",
         "_bluetooth_adapters",
         "_cancel_allocation_callbacks",
         "_cancel_unavailable_tracking",
-        "_clients",
         "_connectable_history",
         "_connectable_scanners",
         "_connectable_unavailable_callbacks",
@@ -174,7 +173,6 @@ class BluetoothManager:
         "_debug",
         "_demoted_sources",
         "_disappeared_callbacks",
-        "_disconnect_tasks",
         "_fallback_intervals",
         "_intervals",
         "_loop",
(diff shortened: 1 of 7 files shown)


rpodgorny:
CI is green apart from the CodSpeed check, which I believe is noise rather than a real regression. Writing down the reasoning so it doesn't have to be re-derived.

It reports one regression and three improvements:

|     | benchmark | change |
| --- | --------- | ------ |
| ❌ | `test_seed_name_cache_cold_first_name` | −7.51% |
| ⚡ | `test_inject_100_raw_unchanged_advertisements` | +7.48% |
| ⚡ | `test_inject_100_rotating_manufacturer_data` | +7.08% |
| ⚡ | `test_inject_100_complex_advertisements` | +5.43% |

The flagged benchmark is:

```python
def run():
    for _ in range(1000):
        manager._name_cache.pop(address, None)
        manager.seed_name_cache(address, name)
```

That exercises `_name_cache` / `seed_name_cache` only. This PR does not touch the name cache or the advertisement path. Its runtime effect on `BluetoothManager` outside `connect()` and scanner unregister is two extra attribute initialisations in `__init__` (one dict, one `WeakSet`), which this loop never reaches. The three benchmarks reported as *faster* are advertisement hot paths this PR does not touch either.

One regression and three improvements of near-identical magnitude in opposite directions looks like run-to-run variance, and the check itself carries the "Different runtime environments detected" warning — expected for a PR from a fork, where base and head measurements come from different runners.

The one mechanism I would take seriously is that adding two attributes to a `cdef class` changes object layout and could shift cache behaviour. But that would be systematic in one direction, not three-up-one-down, and it would not selectively affect a loop that touches neither new attribute.

Happy to drop the `WeakSet`/task-set attributes into a lazily-created holder if you would rather the manager's layout not change at all, though that trades a real allocation on every connect for a hypothetical layout effect.

Since this is a fork PR I cannot acknowledge the result in CodSpeed myself.

bdraco:
@rpodgorny please resolve conflicts

src/habluetooth/manager.py
```diff
@@ -199,6 +203,9 @@ def __init__(
     ) -> None:
         """Init bluetooth manager."""
         self._cancel_unavailable_tracking: asyncio.TimerHandle | None = None
+        # Connected clients per source; weak so a dropped client cannot leak.
+        self._clients: dict[str, WeakSet[HaBleakClientWrapper]] = {}
```
bdraco: this feels like its in the wrong place

bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - A client that reconnected through a different scanner can be spuriously disconnected

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

src/habluetooth/manager.py
bluetoothbot: Agreed. Attribute sits alone above `_advertisement_tracker`, blank-line-separated. Move it beside `_loop` and `_recovery_lock` (line 276). Reorder `manager.pxd` to match. Also rename `_disconnect_tasks` — one producer, disconnects only. Generic name invites unrelated tasks, which would make a blanket `cancel()` in `async_stop()` unsafe later.

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - A client that reconnected through a different scanner can still be spuriously disconnected

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco (approved):

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco (approved):

bluetoothbot (approved):
> [!TIP]

Issue #358, opened 2026-03-08
Requesting a public API to clear per-device advertisement history


FezVrasta:
**Is your feature request related to a problem? Please describe.**

Some BLE devices encode state as mutually-exclusive service UUIDs per advertisement packet, e.g. UUID `0xE800` = state A, UUID `0xE000` = state B. Once HA has seen both states, the merged UUID set stabilises and the change-detection guard in `scanner_adv_received` suppresses all future `async_register_callback` callbacks. The device becomes permanently unresponsive to state changes via passive scanning.

**Describe the solution you'd like**

A public method on `BluetoothManager` to clear the per-device advertisement history for a given address, so the next packet is always treated as new data:

```
def async_clear_advertisement_history(self, address: str) -> None:
    """Clear cached advertisement history for a device.

    Causes the next advertisement from this address to be treated as new
    data, bypassing the change-detection guard. Intended for devices that
    encode state in mutually-exclusive service UUIDs.
    """
    for scanner in self._sources.values():
        scanner._previous_service_info.pop(address, None)

```

Calling this after each advertisement callback keeps callbacks alive indefinitely without touching private internals.

**Additional context**

The workaround currently requires directly popping from `_previous_service_info`, a private field. This works but is fragile. A public API would make this pattern officially supportable. Happy to submit a PR.

PR #613, opened 2026-08-28
fix!: keep the device record when clearing advertisement history

### LINKED ISSUE #612: async_clear_advertisement_history() makes the device unconnectable until it advertises again
### Version

habluetooth 6.26.5 (Home Assistant 2026.8.3)

### The problem

`BluetoothManager.async_clear_advertisement_history()` — added for #358 — also makes the address unconnectable until the device advertises again, because the state it clears is the same state the scanners hand out as their discovered device.

`manager.py:1335`:

```python
def async_clear_advertisement_history(self, address: str) -> None:
    self._all_history.pop(address, None)
    self._connectable_history.pop(address, None)
    ...
    for scanner in self._sources.values():
        scanner._previous_service_info.pop(address, None)   # <-- also the device record
```

`_previous_service_info` does double duty. Two of the three `get_discovered_device_advertisement_data()` implementations read from it:

| implementation | reads from | affected |
| --- | --- | --- |
| `BaseHaRemoteScanner` (`base_scanner.py:804`) | `_previous_service_info` | yes |
| `HaScannerMgmt` (`scanner_mgmt.py:383`) | `_previous_service_info` | yes |
| `HaScannerBleak` (`scanner_bleak.py:290`) | `discovered_devices_and_advertisement_data` | no |

And that is what connecting resolves a backend through:

- `BluetoothManager.async_scanner_devices_by_address()` (`manager.py:462`) builds its list from `scanner.get_discovered_device_advertisement_data(address)`
- `HaBleakClientWrapper._async_get_best_available_backend_and_device()` (`wrappers.py:586`) sorts that list and raises when it is empty

So after calling the public API, `establish_connection()` fails with `No backend ... was found: unknown (never seen by any scanner)` until the next advertisement repopulates the store. ESPHome proxies and the Linux mgmt scanner are affected; the bleak local scanner is not.

### Why this bites the use case the API was added for

#358 is about devices that encode state in mutually-exclusive service UUIDs, where the merged UUID set stabilises and passive callbacks stop. The remedy is to clear the history on **every** advertisement, so each one is treated as new data. Doing that leaves only the microseconds between an advertisement being dispatched and the callback clearing it in which a connection can resolve a backend — so the device becomes effectively unconnectable.

Observed on an ISEO Argo BLE lock (the device from #358), advertising every ~2 s, with the history cleared from the advertisement callback:

```
10:40:48 DEBUG Parsing ISEO advertisement: marker=0xf001, state=0xe800
10:40:51 DEBUG Parsing ISEO advertisement: marker=0xf001, state=0xe800
10:40:53 DEBUG Connecting to 00:15:42:09:1A:1A for Gateway Open
...
bleak_retry_connector.BleakOutOfConnectionSlotsError: 00:15:42:09:1A:1A: Failed to connect
after 9 attempt(s): No backend with an available connection slot that can reach address
00:15:42:09:1A:1A was found: unknown (never seen by any scanner);
4 scanner(s) registered, 4 scanning, 4 connectable
```

Advertisements are plainly arriving throughout, and the device is a few metres away.

### Suggested fix

Separate the two concerns, so clearing the change-detection state does not remove the device record. Options, roughly in order of preference:

1. Keep the discovered-device record in its own mapping, and have `async_clear_advertisement_history()` reset only the merge/change-detection state.
2. Have `async_clear_advertisement_history()` re-seed `_previous_service_info[address]` with the current advertisement rather than popping it, so change detection is defeated while the device stays reachable.
3. If the coupling is deliberate, document the consequence on the public API — the current docstring does not mention that connections break — and offer a narrower call for the #358 use case.

### Workaround

Skip clearing while a connection is in flight, so the record survives long enough for `bleak-retry-connector` to resolve a backend:

```python
if not self._ble_lock.locked():
    async_clear_advertisement_history(hass, address)
```

This works, but it is timing-dependent: any consumer clearing on every advertisement without such a guard will find the device unconnectable, and that is the documented pattern for #358.

FezVrasta:
Fixes #612.

## Problem

`BluetoothManager.async_clear_advertisement_history()` popped each scanner's `_previous_service_info` entry for the address. That mapping does double duty: it is the merge / change-detection state, but it is also the discovered-device record that two of the three `get_discovered_device_advertisement_data()` implementations serve:

| implementation | reads from | affected |
| --- | --- | --- |
| `BaseHaRemoteScanner` | `_previous_service_info` | yes |
| `HaScannerMgmt` | `_previous_service_info` | yes |
| `HaScannerBleak` | `discovered_devices_and_advertisement_data` | no |

That record is how a connection resolves a backend — `async_scanner_devices_by_address()` builds its list from it, and `HaBleakClientWrapper._async_get_best_available_backend_and_device()` raises when that list is empty. So clearing the history left the address unconnectable until it advertised again:

```
bleak_retry_connector.BleakOutOfConnectionSlotsError: 00:15:42:09:1A:1A: Failed to connect
after 9 attempt(s): No backend with an available connection slot that can reach address
00:15:42:09:1A:1A was found: unknown (never seen by any scanner);
4 scanner(s) registered, 4 scanning, 4 connectable
```

This bites exactly the use case the API was added for (#358). Devices that encode state in mutually-exclusive service UUIDs need the history cleared on *every* advertisement, which leaves only the microseconds between an advertisement being dispatched and the callback clearing it in which a connection can resolve a backend. Observed on an ISEO Argo BLE lock advertising every ~2s a few metres from four proxies.

## Fix

Separate the two concerns. Instead of dropping the scanner's record, `_clear_advertisement_merge_history()` replaces it with one that keeps the parts that make the address reachable — `device`, `name`, `rssi`, `time`, `tx_power` — and empties the parts that merging feeds on: `{}` manufacturer data, `{}` service data, `[]` service UUIDs, and `raw = None`.

That is the whole mechanism; no new state and no new branches:

- the discovered-device record stays in place, so the address remains reachable through that scanner while waiting for the next advertisement, and keeps its `BLEDevice` identity across the clear;
- the existing merge branch, given empty previous data, produces the from-scratch result on its own — merging then resumes normally from the next advertisement;
- `raw = None` makes the unchanged-raw short-circuit in `_async_on_raw_advertisement` miss naturally, so it cannot reuse the already-merged data and defeat the reset;
- `info.time` is carried over, so `_async_expire_devices` still drops the record at the original 195s deadline measured from the last genuine detection. Clearing does not extend a record's lifetime.

Both advertisement hot paths are unchanged. A new object is built rather than mutating the stored one, so an `AdvertisementData` already handed to a consumer is not retroactively emptied.

The public behaviour of `async_clear_advertisement_history()` is otherwise unchanged — the manager-side history, name cache, smoothed RSSI, demoted sources and rescue state are still cleared, so the next advertisement is still treated as new data.

### Why `fix!`

The observable contract of a public API method changes: a caller that relied on the clear making a connect fail until the next advertisement now sees it succeed. The new guarantee is documented on the method itself.

The one behavioural delta is bounded: a connect may resolve a backend through a scanner whose last real advertisement predates the clear, for up to the 195s staleness window. That is the same window a device that simply stopped advertising already has today; previously this particular window was zero.

`async_address_reachability_diagnostics()` gained a matching phrase. History-empty with scanner records present was not reachable before this PR, and the connection-intent diagnostics chose their first phrase purely from history — so after a clear it printed `unknown (never seen by any scanner)` and then listed the scanners that see it. It now reports `not in history (advertisement history was cleared)` when scanners still hold the address.

## Tests

- `test_async_clear_advertisement_history_clears_scanner_merging` drives the scanner through the real advertisement path and asserts merging resets for exactly one advertisement, the `BLEDevice` survives, the retained record keeps its original `time`/`rssi`, and the address stays in `async_scanner_devices_by_address()`.
- `test_async_clear_advertisement_history_from_dispatch_callback` reproduces the #358/#612 flow end to end: clear on every advertisement, from inside the dispatch callback. This pins the load-bearing store-before-dispatch order in `_async_on_advertisement_internal` — the reset returns early with no stored record, so swapping those two lines would silently make the fix a no-op for exactly this use case.
- `test_async_clear_advertisement_history_bypasses_raw_shortcut` pins the `raw = None` mechanism; preserving `raw` makes it fail.
- `test_address_reachability_diagnostics_after_clearing_history` covers the new diagnostics phrase.

Each new assertion was checked to fail against the code without the corresponding change.

Full suite passes both pure-Python and Cython-compiled (941 passed).

---

https://claude.ai/code/session_01LqUfoujs2tnVa1EhhgEhB7


FezVrasta:
## CI is red for a reason unrelated to this PR

Every failing job dies in **`Install Dependencies`**, before any habluetooth code compiles and before a single test runs:

```
src/bluetooth_data_tools/gap.py:210:0: Function signature does not match previous declaration
Cython.Compiler.Errors.CompileError: src/bluetooth_data_tools/gap.py
```

`SKIP_CYTHON` / `REQUIRE_CYTHON` are not private to this repo — every dependency built from an sdist that shares our `build_ext.py` pattern reads them too. Setting `REQUIRE_CYTHON` for the whole `poetry install` re-raises a *dependency's* Cython failure instead of letting it fall back to pure Python, and `bluetooth-data-tools` 1.29.21 does not compile under Cython 3.3.0. Hence the clean split: all `skip_cython` jobs pass, all `use_cython` jobs fail.

Fixed in #614 (stacked on #610, which is also needed). Nothing in this PR needs to change — rebasing on those turns it green.

FezVrasta:
Validated on the hardware from #612 and #358 — an ISEO Argo X1R, which encodes door state as mutually-exclusive service UUIDs and is reached through a mix of a local adapter and ESPHome proxies.

Applied to habluetooth 6.26.5 under Home Assistant 2026.8.3 (compiled modules disabled so the patched Python is used).

**Before**, with `async_clear_advertisement_history()` called from the advertisement callback, every connection attempt failed:

```
bleak_retry_connector.BleakOutOfConnectionSlotsError: 00:15:42:09:1A:1A:
Failed to connect after 9 attempt(s): No backend with an available connection slot
that can reach address 00:15:42:09:1A:1A was found: unknown (never seen by any scanner);
4 scanner(s) registered, 4 scanning, 4 connectable
```

— while advertisements were arriving every 2s. The device was being removed from every scanner's discovered-device record microseconds after each advertisement, so a connection could essentially never resolve a backend.

**After**, connections resolve immediately and advertisements keep flowing:

```
12:35:38.783  Connecting to 00:15:42:09:1A:1A for Gateway Open
12:35:38.895  Connected to 00:15:42:09:1A:1A in 0.11s
...
12:39:04.118  advertisement received, state changed to open
```

Both halves hold: door-state changes still dispatch after the merge reset, and the address stays reachable between advertisements. That matters on this device because it advertises in bursts and then goes quiet for 5–27 minutes, so losing the record meant it was unreachable for almost all of that window.

bdraco:
This looks like an undisclosed breaking change labeled as a fix pr

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Observable behaviour of a public API changes without a release-note disclosure

bdraco:
Needs to wait until after a Home Assistant release cycle to merge since we can't shouldn't merge a breaking change during beta

FezVrasta:
@bdraco are we good to go now?

bdraco:
HA window is no longer a blocker, but needs human review

bdraco:
Mostly I haven't had the time to think out who this breaks and what the implications are for people who expected the clear to make it wait for an adv to connect successfully. I think that's only cases where it will now connect instead of failing to connect where the first case might have been intended to fail but that seems like an out there action at a distance case

bluetoothbot (approved):
> [!TIP]

bdraco:
Pushed a simplification of the mechanism. Instead of tracking marked addresses in a new set (with the pxd, slots, expiry and restore bookkeeping and two hot path checks that came with it), the scanner now replaces the stored record with one that keeps the device, rssi, time and tx_power but has empty service_uuids, service_data, manufacturer_data and no raw bytes. The existing merge branch then produces the from scratch result on its own, and the unchanged raw shortcut misses naturally because raw is None. The BLEDevice identity is preserved across the reset, and the advertisement hot paths are untouched.

Tests were folded into two that drive the real ingestion path.

@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
Happy with the design now as it no longer affects the hot path. I think it's mergable once the PR description is updated and the bot suggestions are addressed

FezVrasta:
Description rewritten around the record-replacement mechanism (the old text still described the marker set), and all three suggestions addressed:

1. **`bcb1891`** — `test_async_clear_advertisement_history_from_dispatch_callback` reproduces the #358/#612 flow: clear on every advertisement, from inside the dispatch callback, driving the real ingestion path. It pins the store-before-dispatch order — swapping `base_scanner.py:723-724` makes it fail (verified), where before nothing did.
2. **`6060110`** — the retained-record guarantee and the "does not extend the record's lifetime" bound are now on the `async_clear_advertisement_history` docstring, plus a *Why `fix!`* section in the description.
3. **`bcaf1ee`** — `_append_connection_diagnostics` no longer says "never seen by any scanner" while listing the scanners that see it; with scanner records present and history empty it reports `not in history (advertisement history was cleared)`. Covered by `test_address_reachability_diagnostics_after_clearing_history`.

Left the advertisement intent alone: it already checks `devices` first, so the contradictory combination is not reachable there.

941 passed, pure-Python and Cython-compiled.

bluetoothbot (approved):
> [!TIP]

bdraco (approved):

FezVrasta:
Re-validated on the hardware, against `bcaf1ee` this time — @bluetoothbot was right that the earlier run exercised the marker set, not the record replacement.

Same ISEO Argo X1R (`00:15:42:09:1A:1A`), HA 2026.9.1 on a Pi 4, five scanners (local hci0 + four ESPHome proxies). To make sure nothing masked the bug, I stripped the two workarounds the integration carries for it: it now clears on **every** advertisement (no "skip while a connection is in flight" guard) and never waits for a fresh advertisement before connecting. Poll interval forced to 10s so connects land in the window right after a clear. habluetooth installed pure-Python (no compiler in the core container), stock 6.26.11 was Cython-compiled.

**With this PR — 7/7 polls connected, 0 failures.** Every one started from a genuinely empty manager history:

```
15:15:01.470  [iseo_argo_ble.lock]      Advertisement reports door_closed=True   ← clears
15:15:02.121  [iseo_argo_ble.lock]      BLE device lookup: manager=None          ← history empty
15:15:02.122  [habluetooth.wrappers]    Found 3 connection path(s), preferred order:
                                        nspanel-pro (RSSI=-64) (slots=2/2 free), …
15:15:03.919  [habluetooth.wrappers]    Connected via nspanel-pro
15:15:04.705  [iseo_argo_ble.client]    Session with 00:15:42:09:1A:1A lasted 0.79s
```

20 advertisements, so 20 clears, across the run. `manager=None` on all 7 confirms `async_clear_advertisement_history` is still clearing what it always cleared — only the scanner record survives.

**Control — stock 6.26.11, same patched integration: 0/5 polls connected.** The backend lookup finds nothing and bleak-retry-connector burns every attempt:

```
15:19:21.528  [habluetooth.wrappers]  00:15:42:09:1A:1A: Looking for backend to connect
15:19:25.539  [habluetooth.wrappers]  00:15:42:09:1A:1A: Looking for backend to connect
   … 9 lookups, no "Found N connection path(s)" line at any of them …
15:19:57.553  [iseo_argo_ble.lock]    Lock is unavailable: … Failed to connect after 10 attempt(s):
              No backend with an available connection slot that can reach address 00:15:42:09:1A:1A
              was found: unknown (never seen by any scanner); 5 scanner(s) registered, 5 scanning,
              5 connectable
```

That is #612 verbatim, on a lock advertising every ~2s a few metres from the proxies.

Instance restored: stock 6.26.11 back in place, integration workarounds back, debug logging off.

One thing the run clarified about scope. `async_ble_device_from_address()` is history-based and still returns `None` after a clear — that is unchanged and correct, and it is why the integration keeps its own last-seen `BLEDevice`. What this PR restores is the *backend resolution* path (`async_scanner_devices_by_address` → `_async_get_best_available_backend_and_device`), which is the one that was raising.

FezVrasta:
Thanks @bdraco , are u going to send a PR to ha-core to update the dep or should I do it?

bdraco:
Please do the HA bump. I'm on an airplane for the next few hours and the wifi is junk

FezVrasta:
> Please do the HA bump. I'm on an airplane for the next few hours and the wifi is junk

that's commitment! 😃

PR #614, opened 2026-08-28
fix(ci): scope REQUIRE_CYTHON to our own build

FezVrasta:
Restores a green CI. **Stacked on #610** — its commit is included here because CI cannot go green without it; the diff collapses to my two commits once #610 lands.

## Why CI is red

Every failing job dies in **`Install Dependencies`**, before any habluetooth code compiles and before a single test runs:

```
src/bluetooth_data_tools/gap.py:210:0: Function signature does not match previous declaration
Cython.Compiler.Errors.CompileError: src/bluetooth_data_tools/gap.py
```

`SKIP_CYTHON` / `REQUIRE_CYTHON` are not private to this repo — every dependency built from an sdist that shares our `build_ext.py` pattern reads them, and `bluetooth-data-tools` does:

```python
    except Exception:
        if os.environ.get("REQUIRE_CYTHON"):
            raise
        pass
```

Setting the variable for the whole `poetry install` therefore re-raises a *dependency's* Cython failure instead of letting it fall back to pure Python, which is exactly what that dependency's own `build_ext` intends. That is why the split is so clean: `skip_cython` jobs pass (the dependency skips its Cython build), `use_cython` jobs fail (it is forced to hard-fail).

`bluetooth-data-tools` 1.29.21 does not compile under Cython 3.3.0 — same annotation-vs-`.pxd` class as #610, in their `gap.py`. 1.29.21 is the current release, so there is nothing to bump to, and only one wheel is published, so nearly every job builds it from sdist.

`main` last ran CI before this and its venv cache was warm, which is why the breakage surfaced on a PR first.

## Changes

**1. `fix(ci)`: keep the variables off the dependency install** — install dependencies first without them, then the root project with them, so the flag applies to our extensions only. This is the fix for the current red.

**2. `fix(build)`: bind the rescue request lookup outside the `or` chain** — separate, pre-existing, and not required for CI today. Cython 3.0.12 through 3.1.x cannot prove that a walrus bound inside a short-circuited `or` assigns a typed local (`requests=set` in `auto_scheduler.pxd`) and rejects the later read:

```
src/habluetooth/auto_scheduler.py:1078:55: local variable 'requests' referenced before assignment
src/habluetooth/auto_scheduler.py:1091:33: local variable 'requests' referenced before assignment
```

`build-system.requires` only asks for `Cython>=3`, so a source build against anything in that range fails. Ordering is unchanged — the dict lookup still only runs once `loop` and `_running` have passed. Drop this commit if you would rather keep the PR to the CI fix.

## Verification

Cythonized `TO_CYTHONIZE` against each release, on a clean `origin/main` worktree and on this branch:

| Cython | main | main + #610 | this branch |
| --- | --- | --- | --- |
| 3.0.11 | ok | ok | ok |
| 3.0.12 | fail (walrus) | fail (walrus) | ok |
| 3.1.8 | fail (walrus) | fail (walrus) | ok |
| 3.2.9 | fail (annotations) | ok | ok |
| 3.3.0 | fail (annotations) | ok | ok |

The dependency failure reproduces standalone with no habluetooth code involved — `bluetooth-data-tools` 1.29.21 sdist plus Cython 3.3.0 is enough.

Ran the two-step install locally end to end: dependencies install clean, all six extensions build, `habluetooth.base_scanner` and `habluetooth.auto_scheduler` import from `.so`, and the suite passes compiled (899 passed, 1 skipped).

---

https://claude.ai/code/session_01KBv9s21KpihdvwPnhEwdYG


bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - `annotation_typing: False` is a competing fix that survived the merge — main already solved this differently

PR #616, opened 2026-08-30
build: pin cython below 3.3

bdraco:
Cython 3.3.0 started using subscripted container annotations such as dict[str, int] as real types, so a pure Python def annotated that way no longer matches the plain dict declaration in the pxd and compilation fails with "Signature not compatible with previous declaration" in base_scanner. Cython is only resolved through build-system requires and is not in the lock file, so any CI cell with a cold venv cache now picks up 3.3.0; the macOS 3.12 use_cython job on #611 is the first one to hit it.

This pins Cython below 3.3 until the pxd signatures can be adapted. The same job also fails while building the bluetooth-data-tools 1.29.21 sdist for the same reason, that needs a bluetooth-data-tools release with the pin or wheels, so this alone will not turn that cell green yet.


bdraco:
@bluetoothbot review

PR #617, opened 2026-08-30
fix: build with cython 3.3.0 by reconciling annotations with the pxd declarations

bdraco:
## Summary
Cython 3.3.0 cannot build any source checkout; five of the seven compiled modules fail, so every source build silently falls back to pure Python. Two new 3.3 behaviors cause it: a global annotated with `typing.Final` that is also `cdef` in the `.pxd` crashes the compiler (cython/cython#7942), and a subscripted annotation like `list[str]` or `set[ActiveScanRequest]` is now resolved and rejected as incompatible with the plain `list`, `dict`, `set` in the `.pxd`.

The fix keeps the `.pxd` as the single source of truth, the same approach as python-zeroconf#1820 and aioesphomeapi#1881. `Final` is dropped from the eight globals that have a `.pxd` declaration, and each failing annotation is routed through an alias Cython cannot resolve, the existing `_str = str` idiom plus `_ActiveScanRequest` and `_DueBucket` in `auto_scheduler.py`, so the annotation degrades to the plain container while mypy still sees the full type. No `.pxd` changes; supersedes and closes #610.

## Details
- With Cython 3.2.9 the generated C for every function body is identical to main after normalizing line numbers and the string table; the only additions are the module level alias bindings executed once at import.
- `dict[Any, Any]`, `_str | None`, `_int | None` and classes imported under `TYPE_CHECKING` already pass; only the 12 signatures touched here conflicted.
- CLAUDE.md gains the two rules so this does not regress.

## Test plan
- [x] all seven modules compile with Cython 3.3.0 and 3.2.9
- [x] full pytest suite passes against wheels built by 3.3.0 and by 3.2.9 (7 `.so` each) and in pure Python mode
- [x] pre-commit (ruff, mypy, codespell) passes


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - Reverting #616 wholesale also downgrades the bluetooth-data-tools floor (undoes this branch's own lock bump)

Issue #615, opened 2026-08-30
Passive scanning or_patterns only match General Discoverable FLAGS, making other devices unconnectable


Resend1298:
## Describe the bug

In passive mode, `PASSIVE_SCANNER_ARGS` (`scanner_bleak.py:54-60`) registers a BlueZ `AdvertisementMonitor1` matching only three exact FLAGS values:

```python
# ... The below matcher will match all devices.
or_patterns=[
    OrPattern(0, AdvertisementDataType.FLAGS, b"\x02"),
    OrPattern(0, AdvertisementDataType.FLAGS, b"\x06"),
    OrPattern(0, AdvertisementDataType.FLAGS, b"\x1a"),
]
```

All three set bit 1 (LE General Discoverable). Contrary to the comment, this does not match `0x05` / `0x04` / `0x01` (LE Limited Discoverable), `0x00`, or advertisements that omit the FLAGS structure entirely.

Unmatched devices never enter bleak's `seen_devices`, which is what `get_discovered_device_advertisement_data()` (`scanner_bleak.py:290`) and therefore `async_scanner_devices_by_address()` (`manager.py:474`) read. With an empty list, `_async_get_best_available_backend_and_device()` raises at `wrappers.py:659`:

```
BleakError: No backend with an available connection slot that can reach address
<addr> was found: in connectable history; 1 scanner(s) registered, 1 scanning,
1 connectable
```

Three things make this hard to recognise:

- The error is raised before `scanner._add_connecting()`, so `connect_failed_total` stays 0 and slots read `slots: 5, free: 5, allocated: []`. Nothing is out of slots; the device just has no resolvable path. `bleak_retry_connector` then surfaces it as `BleakOutOfConnectionSlotsError`, pointing at the wrong cause.
- With the MGMT side channel up, the scanner passes `detection_callback=None` to bleak (`scanner_bleak.py:415-420`) and advertisements reach the manager from the kernel, unfiltered. So state updates keep working normally while every connection fails, and the device looks healthy.
[...]

PR #618, opened 2026-08-30
chore: add advertisement monitor probe script

bdraco:
## Summary
Standalone probe for #615. It registers BlueZ `AdvertisementMonitor1` objects the same way bleak does and records which devices each one reports, so we can see on a real adapter which FLAGS values are in range, whether a single monitor silently drops patterns past the kernel limit of 16 (`HCI_MAX_ADV_MONITOR_NUM_PATTERNS`), whether two monitors of 16 cover everything, and how much the current three value list misses. That decides the shape of the fix.

Needs only `dbus_fast`, BlueZ 5.56+ with `--experimental` and kernel 5.10+; it never starts discovery or connects.

```
python scripts/probe_adv_monitor.py --adapter hci0 --duration 20
```

Six phases of 20s by default (four census batches, the legacy list, one 32 pattern monitor, two 16 pattern monitors), then a per value table and totals. `--skip-census` runs only the last three.

## Test plan
- [x] imports and `--help` work with the repo venv
- [ ] run on a BlueZ host with a FLAGS 0x05 device in range


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

PR #31, opened 2024-02-29
fix: android beacons in passive mode with flags 0x02

Sorien:
bleak uses DBus BlueZ Interface org.bluez.AdvertisementMonitor1 that will return ble packet data always filtred by [data section flags](https://www.google.com/search?q=ble+datasection+flags&oq=ble+datasection+flags&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIGCAEQLhhA0gEINDA1NGowajSoAgCwAgA&sourceid=chrome&ie=UTF-8#ip=1) 
```
BLE_SIG_ADV_FLAG_LE_LIMITED_DISC_MODE (0x01)
BLE_SIG_ADV_FLAG_LE_GENERAL_DISC_MODE (0x02)
BLE_SIG_ADV_FLAG_BR_EDR_NOT_SUPPORTED (0x04)
BLE_SIG_ADV_FLAG_LE_BR_EDR_CONTROLLER (0x08)
BLE_SIG_ADV_FLAG_LE_BR_EDR_HOST (0x10)
BLE_SIG_ADV_FLAGS_LE_ONLY_LIMITED_DISC_MODE (0x05)
BLE_SIG_ADV_FLAGS_LE_ONLY_GENERAL_DISC_MODE (0x06)
```
there seems to be no way to force it to return all packets like in active scan mode 

currenty we are filterining flags **06** `(LE_SIG_ADV_FLAG_LE_GENERAL_DISC_MODE and BLE_SIG_ADV_FLAG_BR_EDR_NOT_SUPPORTED)` or **1a** `(LE_SIG_ADV_FLAG_LE_GENERAL_DISC_MODE, BLE_SIG_ADV_FLAG_LE_BR_EDR_HOST, BLE_SIG_ADV_FLAG_LE_BR_EDR_CONTROLLER)`

problem is that Android BLE doesn't have API to configure these flags, there is one solution we can set [AdvertiseSettings.Builder.setConnectable](https://developer.android.com/reference/android/bluetooth/le/AdvertiseSettings.Builder#setConnectable(boolean)) to true and we will get 
image

it should be already set in HA companion app

but as you can see it sets just flag BLE_SIG_ADV_FLAG_LE_GENERAL_DISC_MODE and currently such packet will be filtered out.

this PR changes filtering to include all ble packets that have set BLE_SIG_ADV_FLAG_LE_GENERAL_DISC_MODE (0x02)

all BLE devices that was working before this change continues working 
IBeacon integration works in passive mode for Android devices

image
image

PR #619, opened 2026-08-30
fix: passive scanning misses devices whose flags are not general discoverable

### LINKED ISSUE #615: Passive scanning or_patterns only match General Discoverable FLAGS, making other devices unconnectable

bdraco:
Fixes #615.

The passive or_patterns only matched FLAGS 0x02, 0x06 and 0x1a, so devices advertising 0x04, 0x05 (SwitchBot Bot) or 0x18 never reached the discovered map the connect path reads and every connect failed. The list now holds the values seen in the wild (#31 and the probe census from #618); a monitor takes at most 16 patterns and BlueZ releases larger ones, a test guards that.

- [x] tests, full suite passes
- [x] manual check with a FLAGS 0x05 device in AUTO mode

### DIFF
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1554,14 +1554,6 @@ class BluetoothManager:
         else:
             parts.append("unknown (never seen by any scanner)")
 
-        # History is fed but no scanner has the device cached (passive
-        # or_patterns miss); not an out of slots condition.
-        if not devices and address in self._all_history:
-            parts.append(
-                "in history but no scanner currently has it in its discovered devices"
-            )
-            return
-
         connectable_devices = [d for d in devices if d.scanner.connectable]
         non_connectable_devices = [d for d in devices if not d.scanner.connectable]
         if not connectable_devices and non_connectable_devices:
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1554,6 +1554,16 @@ class BluetoothManager:
         else:
             parts.append("unknown (never seen by any scanner)")
 
+        # A local scanner can keep feeding history through the raw
+        # advertisement channel while its bleak discovered map never sees
+        # the device (passive or_patterns), so say so instead of letting
+        # the message read as an out of slots condition.
+        if not devices and address in self._all_history:
+            parts.append(
+                "in history but no scanner currently has it in its discovered devices"
+            )
+            return
+
         connectable_devices = [d for d in devices if d.scanner.connectable]
         non_connectable_devices = [d for d in devices if not d.scanner.connectable]
         if not connectable_devices and non_connectable_devices:
--- a/src/habluetooth/manager.py
+++ b/src/habluetooth/manager.py
@@ -1554,10 +1554,8 @@ class BluetoothManager:
         else:
             parts.append("unknown (never seen by any scanner)")
 
-        # A local scanner can keep feeding history through the raw
-        # advertisement channel while its bleak discovered map never sees
-        # the device (passive or_patterns), so say so instead of letting
-        # the message read as an out of slots condition.
+        # History is fed but no scanner has the device cached (passive
+        # or_patterns miss); not an out of slots condition.
         if not devices and address in self._all_history:
             parts.append(
                 "in history but no scanner currently has it in its discovered devices"
--- a/src/habluetooth/scanner_bleak.py
+++ b/src/habluetooth/scanner_bleak.py
@@ -48,15 +48,9 @@ if IS_LINUX:
     from dbus_fast import InvalidMessageError
     from dbus_fast.service import method
 
-    # BlueZ passive scanning only creates and updates a Device1 object for
-    # advertisements that match a registered or_pattern, and the only AD
-    # structure common to connectable advertisements is FLAGS, so the
-    # monitor lists FLAGS values. A monitor holds at most 16 patterns
-    # (kernel HCI_MAX_ADV_MONITOR_NUM_PATTERNS) and BlueZ releases a
-    # monitor with more instead of truncating it. These are the values
-    # observed on real networks: 0x02 android beacons (#31), 0x04 and
-    # 0x18 non discoverable connectable devices, 0x05 limited discoverable
-    # (SwitchBot bots), 0x06 and 0x1a general discoverable (#615).
+    # Passive scanning only reports devices matching an or_pattern; a
+    # monitor holds at most 16 (BlueZ releases larger ones). FLAGS values
+    # seen in the wild: #31 (0x02) and #615.
     PASSIVE_SCAN_FLAGS = (0x02, 0x04, 0x05, 0x06, 0x18, 0x1A)
     PASSIVE_SCANNER_ARGS = BlueZScannerArgs(
         or_patterns=[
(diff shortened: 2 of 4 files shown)


bdraco:
@bluetoothbot review

bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bdraco:
@bluetoothbot review

bdraco:
> 2. Allowlist can never be exhaustive, and 10 of the 16 pattern slots are left unused
> [src/habluetooth/scanner_bleak.py:51-54](https://github.com/Bluetooth-Devices/habluetooth/blob/c2454cf284ab983ef5fd172438635f0992cb9e84/src/habluetooth/scanner_bleak.py#L51-L54)

accepted and I have over 150 test bluetooth devices showing its likely unneeded to expand. not going to expand it unless we have proof its needed and each slot used is one we can't use later

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

PR #620, opened 2026-08-30
chore: make the probe survive a refused monitor and gate the overflow test

bdraco:
## Summary
Follow up to #618, pushed a minute after it merged. The run on a BlueZ host showed that a monitor with more than 16 patterns is never activated and is released by bluetoothd, and the bus connection dropped right after, which took the two monitor phase down with it.

The probe now treats a refused RegisterMonitor as a result instead of a crash, runs the two monitor test before anything risky, and only registers the 32 pattern monitor with `--test-overflow`, last.

## Test plan
- [x] `--help` and import with the repo venv

### DIFF
--- a/scripts/probe_adv_monitor.py
+++ b/scripts/probe_adv_monitor.py
@@ -11,9 +11,10 @@ first, export the object second) and records which devices each monitor
 reports through DeviceFound, then prints a table answering:
 
 1. which FLAGS byte values the devices in range actually advertise;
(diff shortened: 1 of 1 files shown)


bdraco:
@bluetoothbot review

bluetoothbot (changes requested):
> [!WARNING]
> Important issues found.
>
> - unregister() pairs the wrong monitors with paths — crashes in the exact case the PR is about

PR #622, opened 2026-08-30
fix: say when a device is in history but cached by no scanner

bdraco:
When a device is in history but no scanner has it in its discovered map, the connect error read as an out of slots condition. The connect intent diagnostics now say `in history but no scanner currently has it in its discovered devices`. Found while working on #615.

- [x] test, full suite passes


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

PR #624, opened 2026-08-30
fix: restore discoveries after an active window flip

bdraco:
bleak 3 clears `seen_devices` on every `start()` and the in place active window flip skipped the re-seed a full start does, so every AUTO flip emptied the map the connect path reads until devices were rediscovered. Re-seed with `restore_discoveries_sync` (bleak-retry-connector 4.7.0) right after the restart; it never awaits, so nothing runs between the restart and the re-seed. Found while working on #615. Stacked on the 4.7.0 bump PR.

- [x] test asserts one restore per toggle, full suite passes


bdraco:
@bluetoothbot review

bluetoothbot (approved):
> [!TIP]

bluetoothbot (approved):
> [!TIP]

PR #631, opened 2026-09-08
fix(manager): dispatch changed advertisements from non-owner scanners

rankjie:
I ran into this with a Xiaomi presence sensor: nine scanners received its clear event, but the preferred scanner missed it. Source arbitration dropped the new advertisement before checking its payload, leaving the Home Assistant entity occupied. The trace is described in https://github.com/home-assistant/core/issues/179913#issuecomment-5474565917.

This lets changed advertisements from non-preferred scanners reach integration callbacks while keeping the existing source selection and connection routing. A separate last-dispatched record avoids repeating identical updates from other non-preferred scanners and is cleared when the device disappears or its history is reset.

I've deployed this as a compiled wheel on Home Assistant Core 2026.9.1 with habluetooth 6.26.11. Installation, rollback and startup were verified; longer-term missed-event monitoring is still ongoing.

Related to https://github.com/home-assistant/core/issues/179913.


rankjie:
Added coverage for the missing connectable-promotion path, along with owner preservation and duplicate suppression. The three cases fail on the base branch and pass with this change.

bdraco (changes requested):
This PR introduces a major correctness problem. It will allow advertisements from scanners that are out of sync to reach callbacks breaking the ordering guarantee.

bluetoothbot (changes requested):
> [!CAUTION]
> Critical issues found.
>
> - Non-owner dispatch has no ordering/staleness guard and can pin state to a stale payload
> - Bleak callbacks are skipped on the new dispatch path
> - Dispatched payload contradicts async_last_service_info / discovered_service_info

